PR #6, opened 2020-09-09
Feature/new firmware refactor clean raw conversion

cdoehn:
* Use templated functions for better readibility

### DIFF
--- a/src/start_request.cpp
+++ b/src/start_request.cpp
@@ -54,39 +54,39 @@ StartRequest::RawType StartRequest::toRawType() const
 {
   std::ostringstream os;
 
-  write<uint32_t>(os, crc_);
-  write<uint32_t>(os, seq_number_);
-  write<uint64_t>(os, RESERVED_);
-  write<uint32_t>(os, OPCODE_);
+  write(os, crc_);
+  write(os, seq_number_);
+  write(os, RESERVED_);
+  write(os, OPCODE_);
 
   uint32_t host_ip_big_endian = htobe32(host_ip_);
-  write<uint32_t>(os, host_ip_big_endian);
+  write(os, host_ip_big_endian);
 
-  write<uint16_t>(os, host_udp_port_data_);
-  write<uint8_t>(os, device_enabled_);
-  write<uint8_t>(os, intensity_enabled_);
-  write<uint8_t>(os, point_in_safety_enabled_);
-  write<uint8_t>(os, active_zone_set_enabled_);
-  write<uint8_t>(os, io_pin_enabled_);
-  write<uint8_t>(os, scan_counter_enabled_);
-  write<uint8_t>(os, speed_encoder_enabled_);
-  write<uint8_t>(os, diagnostics_enabled_);
+  write(os, host_udp_port_data_);
+  write(os, device_enabled_);
+  write(os, intensity_enabled_);
+  write(os, point_in_safety_enabled_);
+  write(os, active_zone_set_enabled_);
+  write(os, io_pin_enabled_);
+  write(os, scan_counter_enabled_);
+  write(os, speed_encoder_enabled_);
+  write(os, diagnostics_enabled_);
 
   uint16_t start_angle{ radToTenthDegree(master_.getStartAngle()) };
   uint16_t end_angle{ radToTenthDegree(master_.getEndAngle()) };
   uint16_t resolution{ radToTenthDegree(master_.getResolution()) };
-  write<uint16_t>(os, start_angle);
-  write<uint16_t>(os, end_angle);
-  write<uint16_t>(os, resolution);
+  write(os, start_angle);
+  write(os, end_angle);
+  write(os, resolution);
 
   for (const auto& slave : slaves_)
   {
     uint16_t slave_start_angle{ radToTenthDegree(slave.getStartAngle()) };
     uint16_t slave_end_angle{ radToTenthDegree(slave.getEndAngle()) };
     uint16_t slave_resolution{ radToTenthDegree(slave.getResolution()) };
-    write<uint16_t>(os, slave_start_angle);
-    write<uint16_t>(os, slave_end_angle);
-    write<uint16_t>(os, slave_resolution);
+    write(os, slave_start_angle);
+    write(os, slave_end_angle);
+    write(os, slave_resolution);
   }
 
   std::string data_str(os.str());
@@ -102,7 +102,7 @@ StartRequest::RawType StartRequest::toRawType() const
 template <typename T>
 void StartRequest::write(std::ostringstream& os, const T& data) const
 {
-  os.write((char*)&data, sizeof(T));
+  os.write((char*)(&data), sizeof(T));
   return;
 }
 
--- a/src/start_request.cpp
+++ b/src/start_request.cpp
@@ -54,39 +54,39 @@ StartRequest::RawType StartRequest::toRawType() const
 {
   std::ostringstream os;
 
-  os.write((char*)&crc_, sizeof(crc_));
-  os.write((char*)&seq_number_, sizeof(seq_number_));
-  os.write((char*)&RESERVED_, sizeof(RESERVED_));
-  os.write((char*)&OPCODE_, sizeof(OPCODE_));
+  write<uint32_t>(os, crc_);
+  write<uint32_t>(os, seq_number_);
+  write<uint64_t>(os, RESERVED_);
+  write<uint32_t>(os, OPCODE_);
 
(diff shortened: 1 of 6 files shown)


include/psen_scan_v2/reply_msg_from_scanner.h
```diff
@@ -41,23 +41,30 @@ class ReplyMsgFromScanner
 public:
   static ReplyMsgFromScanner fromRawData(const RawScannerData& data);
 
-public:
-  ReplyMsgFromScanner(const uint32_t op_code, const uint32_t res_code);
+private:
```
hslusarek, line 44: Please, try to stick to a certain order of the access modifiers for easier readability:
1. First, all public methods. This is the most important point. This allows the reader to overview the interface of the class quickly. 
2. Then all private methods.
3. Then all private member, etc.

include/psen_scan_v2/reply_msg_from_scanner.h
```diff
@@ -78,12 +84,19 @@ inline uint32_t ReplyMsgFromScanner::calcCRC(const ReplyMsgFromScanner& msg)
...
+  process_bytes<uint32_t>(result, msg.opcode_);
+  process_bytes<uint32_t>(result, msg.res_code_);
   return result.checksum();
 }
 
+template <typename T>
+inline void ReplyMsgFromScanner::process_bytes(boost::crc_32_type& crc_32, const T& data)
+{
+  crc_32.process_bytes(&(data), sizeof(T));
+  return;
```
hslusarek, line 97: Obsolete -> remove.

include/psen_scan_v2/reply_msg_from_scanner.h
```diff
@@ -41,23 +41,30 @@ class ReplyMsgFromScanner
...
+  ReplyMsgFromScanner(const uint32_t op_code, const uint32_t res_code);
   ReplyMsgFromScannerType type() const;
-
-public:
   static uint32_t getStartOpCode();
   static uint32_t calcCRC(const ReplyMsgFromScanner& msg);
 
+private:
+  template <typename T>
+  static void process_bytes(boost::crc_32_type& crc_32, const T& data);
```
hslusarek, line 56: The naming convention for functions is camel case see [wiki.ros.org](http://wiki.ros.org/CppStyleGuide#Function_.2F_Methods).

src/start_request.cpp
```diff
@@ -54,39 +54,39 @@ StartRequest::RawType StartRequest::toRawType() const
 {
   std::ostringstream os;
 
-  os.write((char*)&crc_, sizeof(crc_));
-  os.write((char*)&seq_number_, sizeof(seq_number_));
-  os.write((char*)&RESERVED_, sizeof(RESERVED_));
-  os.write((char*)&OPCODE_, sizeof(OPCODE_));
+  write<uint32_t>(os, crc_);
```
hslusarek, line 57: You can skip the template type here. The type is automatically determined by the compiler.

include/psen_scan_v2/reply_msg_from_scanner.h
```diff
@@ -132,21 +152,29 @@ inline ReplyMsgFromScanner::RawType ReplyMsgFromScanner::toCharArray()
   std::ostringstream os;
 
   uint32_t crc{ calcCRC(*this) };
-  os.write((char*)&crc, sizeof(uint32_t));
-  os.write((char*)&reserved_, sizeof(uint32_t));
-  os.write((char*)&opcode_, sizeof(uint32_t));
-  os.write((char*)&res_code_, sizeof(uint32_t));
 
-  ReplyMsgFromScanner::RawType ret_val{};
+  write<uint32_t>(os, crc);
```
hslusarek, line 156: You can skip the template type here (and in the following lines). The type is automatically determined by the compiler.

include/psen_scan_v2/reply_msg_from_scanner.h
```diff
@@ -99,16 +112,12 @@ inline ReplyMsgFromScanner ReplyMsgFromScanner::fromRawData(const RawScannerData
...
-  // boost::iostreams::stream<Device> stream((char*)&data, sizeof(DataReply::MemoryFormat));
-
-  std::istringstream stream(std::string((char*)&data, REPLY_MSG_FROM_SCANNER_SIZE));
+  std::istringstream is(std::string((char*)&data, REPLY_MSG_FROM_SCANNER_SIZE));
 
-  stream.read((char*)&msg.crc_, sizeof(ReplyMsgFromScanner::crc_));
-  stream.read((char*)&msg.reserved_, sizeof(ReplyMsgFromScanner::reserved_));
-  stream.read((char*)&msg.opcode_, sizeof(ReplyMsgFromScanner::opcode_));
-  stream.read((char*)&msg.res_code_, sizeof(ReplyMsgFromScanner::res_code_));
+  read<uint32_t>(is, msg.crc_);
```
hslusarek, line 117: You can skip the template type here (and in the following lines). The type is automatically determined by the compiler.

include/psen_scan_v2/reply_msg_from_scanner.h
```diff
@@ -69,7 +76,6 @@ class ReplyMsgFromScanner
   //! If the CRC is not correct, the device will not send any message.
   uint32_t res_code_{ 0 };
 
-private:
```
hslusarek, line 72: Why did you delete these modifiers? Of course they are obsolete in the sense that the modifier is already stated above, but they increase the readability and provide a visual separator for the reader of the class.

include/psen_scan_v2/reply_msg_from_scanner.h
```diff
@@ -41,23 +41,30 @@ class ReplyMsgFromScanner
...
   using RawType = std::array<char, REPLY_MSG_FROM_SCANNER_SIZE>;
   RawType toCharArray();
 
 private:
+  template <typename T>
+  void write(std::ostringstream& os, const T& data) const;
+
   ReplyMsgFromScanner() = delete;
 
-private:
```
hslusarek, line 60: Why did you delete these modifiers? Of course they are obsolete in the sense that the modifier is already stated above, but they increase the readability and provide a visual separator for the reader of the class.

include/psen_scan_v2/reply_msg_from_scanner.h
hslusarek, line 56: You are "misleading" the reader/user of the function here because as stated in the [docu](https://en.cppreference.com/w/cpp/io/basic_istream/read) of the `std::istringstream::read()` function, it does not guarantee that the `data` argument is not changed (because the `data` argument is not defined as `const char*` argument). 
Your code only works because a C-Cast is used here which not only casts to a different pointer type but also casts away the constness of the `data` argument. If you use a `reinterpret_cast`, the compiler will complain about this fact.
My suggestion is: Remove the `const` from the `data` argument and use the `reinterpret_cast` instead of the C-cast.

include/psen_scan_v2/reply_msg_from_scanner.h
```diff
@@ -41,23 +41,30 @@ class ReplyMsgFromScanner
...
+  template <typename T>
+  static void process_bytes(boost::crc_32_type& crc_32, const T& data);
+
+public:
   using RawType = std::array<char, REPLY_MSG_FROM_SCANNER_SIZE>;
   RawType toCharArray();
 
 private:
+  template <typename T>
+  void write(std::ostringstream& os, const T& data) const;
```
hslusarek, line 64: I noticed that the `write()` method is basically the same in the `ReplyMsgFromScanner` class and in the `StartRequest` class. Therefore, I suggest to separate this function from the before mentioned classes and, consequently, avoid code duplication.

include/psen_scan_v2/reply_msg_from_scanner.h
```diff
@@ -99,16 +112,12 @@ inline ReplyMsgFromScanner ReplyMsgFromScanner::fromRawData(const RawScannerData
 {
   ReplyMsgFromScanner msg{ 0, 0 };
 
-  // Alternatives for transformation:
-  // typedef boost::iostreams::basic_array_source<char> Device;
-  // boost::iostreams::stream<Device> stream((char*)&data, sizeof(DataReply::MemoryFormat));
-
-  std::istringstream stream(std::string((char*)&data, REPLY_MSG_FROM_SCANNER_SIZE));
+  std::istringstream is(std::string((char*)&data, REPLY_MSG_FROM_SCANNER_SIZE));
```
hslusarek, line 115: Instead of `(char*)&data` use `const_cast<char*>(data.data())`

PR #9, opened 2020-09-11
Feature/readme and x axis rotation

jschleicher:
Simplify launch parameters, add Migration section to Readme

* add steps for migration from former ROS package to Readme
* add supported firmware version to readme
* deprecate x_axis_rotation, set x-axis as zero-direction

Please double check the implementation! I don't know, where best to add the x-axis-offset. As a quick suggestion I've added the offset before the ScannerConfiguration constructor, so that the `configuration` still has zero on the left and only the outermost layer has ROS "zero towards tf x-axis" convention. Tested with real scanner: Still receiving data.


launch/psen_scan_v2.launch
```diff
@@ -33,23 +33,19 @@ along with this program.  If not, see <https://www.gnu.org/licenses/>.
...
   <node name="laser_scanner" type="psen_scan_v2_node" pkg="psen_scan_v2" output="screen" required="true">
-    <rosparam param="host_ip" subst_value="True">$(arg host_ip)</rosparam>
-    <rosparam param="host_udp_port_data" subst_value="True">$(arg host_udp_port_data)</rosparam>
-    <rosparam param="host_udp_port_control" subst_value="True">$(arg host_udp_port_control)</rosparam>
-    <rosparam param="sensor_ip" subst_value="True">$(arg sensor_ip)</rosparam>
-    <rosparam param="frame_id" subst_value="True">$(arg frame_id)</rosparam>
-    <rosparam param="angle_start" subst_value="True">$(arg angle_start)</rosparam>
-    <rosparam param="angle_end" subst_value="True">$(arg angle_end)</rosparam>
-    <rosparam param="x_axis_rotation" subst_value="True">$(arg x_axis_rotation)</rosparam>
+    <param name="host_ip" value="$(arg host_ip)" />
```
ct2034, line 42: What is this change doing?

ct2034 (approved):
Please answer my comment. Afterwards it is good to merge.

launch/psen_scan_v2.launch
jschleicher: The `rosparam` syntax requires an additional `subst_value="True"` argument, to evaluate variables, the `param` syntax is shorter (and thus better imho). Do you see any advantages of the `rosparam` syntax? The only advantage is, that you can set multiple parameters / arrays as yaml inside `rosparam`, but we do not use that here...

agutenkunst:
Please hold this PR for a moment. I am currently trying to figure out why the scan seems to be reversed...

agutenkunst:
I rebased this and assign it to me. There are a few considerations to make. I will check for it next week.

agutenkunst (changes requested):
We will have to discuss where the x_axis_rotation is valid at all. Mixing "frames" seems dangerous to me.

agutenkunst:
As soon as https://github.com/PilzDE/psen_scan_v2/tree/feature/new_firmware_single_monitoring_frame is green I merge this

agutenkunst (approved):

PR #10, opened 2020-09-11
Add stop functionality from old repository psen_scan

hslusarek:


include/psen_scan_v2/scanner_controller.h
```diff
@@ -64,14 +66,16 @@ class ScannerControllerT
   FRIEND_TEST(ScannerControllerTest, test_start_method_calls_correct_state_machine_event);
   FRIEND_TEST(ScannerControllerTest, test_stop_method_calls_correct_state_machine_event);
   FRIEND_TEST(ScannerControllerTest, test_udp_clients_listen_before_sending_start_request);
+  FRIEND_TEST(ScannerControllerTest, testStopRequestSending);
```
agutenkunst, line 69: Try to match the naming scheme to the tests above

include/psen_scan_v2/stop_request.h
```diff
@@ -0,0 +1,53 @@
...
+  //! @brief Serializes the request into raw data which can be send to the scanner.
+  DynamicSizeRawData toRawData() const;
+
+private:
+  std::ostringstream& processMember(std::ostringstream& os) const;
+  uint32_t calcCrc() const;
+
+private:
+  const std::array<uint8_t, NUM_RESERVED_FIELDS> RESERVED_{};
+  const uint32_t OPCODE_{ htole32(0x36) };
```
agutenkunst, line 48: Do we need htole32?

agutenkunst (changes requested):
Just minor finding. Approve after check on real system.

include/psen_scan_v2/scanner_controller.h
hslusarek: Hmm, but I thought we use camel case. Therefore, I suggest changing the existing tests.

include/psen_scan_v2/stop_request.h
hslusarek: StartRequest also uses `htole32()`. But I'm not sure about that...

include/psen_scan_v2/scanner_controller.h
hslusarek: Done.

include/psen_scan_v2/stop_request.h
hslusarek: The ensures that  little-endian is send. But maybe we can store the OP_Code as 0x36 and call the function when the `StopRequest` is converter to raw data.

PR #13, opened 2020-09-15
Extends the code to ensure proper shutdown

hslusarek:
This PR adds ensures that the system is properly shutdown in case the ROS node is shutdown via `Ctrl+C`.


include/psen_scan_v2/udp_connection_state_machine.h
```diff
@@ -144,11 +169,13 @@ struct udp_connection_state_machine_ : public msm::front::state_machine_def<udp_
...
-    a_row < s::init,                       e::start_request,                 s::wait_for_start_reply,      &m::action_send_start_request   >,
+    a_row < s::idle,                       e::start_request,                 s::wait_for_start_reply,      &m::action_send_start_request   >,
+    a_row < s::idle,                       e::stop_request,                  s::wait_for_stop_reply,       &m::action_send_stop_request    >,
      _row < s::wait_for_start_reply,       e::start_reply_received,          s::wait_for_monitoring_frame                                  >,
      _row < s::wait_for_monitoring_frame,  e::monitoring_frame_received,     s::wait_for_monitoring_frame                                  >,
-    a_row < s::wait_for_monitoring_frame,  e::stop_request,                  s::wait_for_stop_reply,      &m::action_send_stop_request     >,
-     _row < s::wait_for_stop_reply,        e::stop_reply_received,           s::init                                                       >
+    a_row < s::wait_for_start_reply,       e::stop_request,                  s::wait_for_stop_reply,       &m::action_send_stop_request    >,
```
hslusarek, line 176: @agutenkunst  Did we test this case? What happens if I send a `StartRequest` and before the `StartReply` is received I send a `StopRequest`? Do we and the device properly handle this case?

PR #18, opened 2020-09-21
Feature/new firmware single monitoring frame

agutenkunst:
Creating draft for discussion


include/psen_scan_v2/laserscan_conversions.h
```diff
@@ -0,0 +1,40 @@
...
+#include "psen_scan_v2/monitoring_frame_msg.h"
+
+namespace psen_scan_v2
+{
+LaserScan toLaserScan(const MonitoringFrameMsg& frame)
+{
+  const double resolution = tenthDegreeToRad(frame.resolution());
+  const double min_angle = tenthDegreeToRad(frame.fromTheta());
+  const uint16_t number_of_samples = frame.measures().size();
+  const double max_angle = min_angle + resolution * number_of_samples;
```
agutenkunst, line 30:
```suggestion
  const double max_angle = min_angle + resolution * (number_of_samples - 1);
```

should be correct right?

src/monitoring_frame_msg.cpp
```diff
@@ -0,0 +1,122 @@
...
+
+  for (unsigned i = 0; i < number_of_samples; i++)
+  {
+    uint16_t sample;
+    raw_processing::read(is, sample);
+    measures_.at(i) = sample / 1000.;  // Convert to mm
+  }
+}
+
+void MonitoringFrameMsg::checkFixedFields()
```
agutenkunst, line 99: @martiniil @cdoehn 
Why are these checks to strict? I think we really should be robust on the data we receive...

src/monitoring_frame_msg.cpp
```diff
@@ -0,0 +1,122 @@
...
+  {
+    msg.deserializeAdditionalField(is);
+  }
+
+  return msg;
+}
+
+void MonitoringFrameMsg::deserializeAdditionalField(std::istringstream& is)
+{
+  const FieldHeader header(is);
```
agutenkunst, line 68: So there is `raw_processing::read` 

I suggest that you keep the complexity low by avoiding classes that take the stream (and change it of course) in order to construct themselfs. Maybe `raw_processing::readHeader` would be a feasible solution?

include/psen_scan_v2/monitoring_frame_msg.h
```diff
@@ -0,0 +1,156 @@
...
+static constexpr uint32_t MAX_SCANNER_ID{ 0x03 };
+
+class FieldHeader
+{
+public:
+  using Id = uint8_t;
+  using Length = uint16_t;
+
+public:
+  FieldHeader(std::istringstream& is);
```
agutenkunst, line 44: I very much dislike this construction. It also adds so much complexity that `length_` could end up being 65535. However this would clearly be suspicious.

include/psen_scan_v2/raw_processing.h
```diff
@@ -26,6 +28,16 @@ inline void write(std::ostringstream& os, const T& data)
 {
   os.write(reinterpret_cast<const char*>(&data), sizeof(T));
 }
+
+template <typename T>
+inline void read(std::istringstream& is, T& data)
+{
+  is.read(reinterpret_cast<char*>(&data), sizeof(T));
```
agutenkunst, line 35: I think here should be some checks in place...

include/psen_scan_v2/laserscan_conversions.h
martiniil: No, see protocol, p.8

src/monitoring_frame_msg.cpp
```diff
@@ -0,0 +1,122 @@
...
+constexpr FieldHeader::Id AdditionalFieldIds::END_OF_FRAME;
+
+FieldHeader::FieldHeader(std::istringstream& is)
+{
+  raw_processing::read(is, id_);
+  raw_processing::read(is, length_);
+  length_--;
+}
+
+MonitoringFrameMsg MonitoringFrameMsg::fromRawData(const MaxSizeRawData& data)
```
agutenkunst, line 41: I think this function could be much simple by relying on `raw_processing::read` and maybe some variants of it.
`deserializeAdditionalField` could/should be simplified by a simple loop + switch case. As long as we are doing simple deserialization there is no benefit is adding a strategy pattern. I'd rather argument that this is more error prone due to the complexity of passing a changing stream around...

src/monitoring_frame_msg.cpp
martiniil: How about `std::pair<uint8_t, uint16_t> MonitoringFrameMsg::readFieldHeader(std::istringstream&)`?

include/psen_scan_v2/monitoring_frame_msg.h
martiniil: See above discussion

src/monitoring_frame_msg.cpp
cdoehn: I changed the exceptions to log messages. The only issue is that they might spam the system if there are too much

src/start_request.cpp
```diff
@@ -22,14 +22,13 @@
...
-#include "psen_scan_v2/tenth_degree_conversion.h"
+#include "psen_scan_v2/angle_conversions.h"
 #include "psen_scan_v2/start_request.h"
-#include "psen_scan_v2/degree_to_rad.h"
 #include "psen_scan_v2/raw_processing.h"
 
 namespace psen_scan_v2
 {
-static constexpr double MASTER_RESOLUTION_RAD{ degreeToRad(1.) };
+static constexpr double MASTER_RESOLUTION_RAD{ degreeToRadian(1.) };
```
cdoehn, line 31: The resolution of the master is 0.1 degree, isn't it?

martiniil: Where is the resolution defined?

agutenkunst: Ok now I get it. Yes we could set 0.1 here I think...

cdoehn (approved):

PR #20, opened 2020-09-22
Use internal transition in state machine

agutenkunst:
Will not call on_entry and on_exit of the state which caused trouble
  using the rosconsole_bridge. The whole ros system would get
  unresponsive.

Please merge before https://github.com/PilzDE/psen_scan_v2/pull/18


martiniil (approved):
:+1:

PR #22, opened 2020-09-22
Adds diagram for calls and callbacks

rfeistenauer:


agutenkunst (changes requested):
Some findings:
- Please try to rectify the lines where possible and check that leveling is always the same
- Maybe add background color to the class names
- Add information what the colors mean
- Switch the yellow to some blue for more accessibility
- I suggest removing the if "Start" and "Stop" since this visualizes a certain case

agutenkunst:
Also I suggest adding lines for composition between the classes according to UML style

agutenkunst (approved):

PR #23, opened 2020-09-22
Add Scanner-API integrationtest

hslusarek:
This PR adds an integration-test which uses all "real" software components of the overall system and simulates the real hardware scanner with a mocked UDP server.

**Note:**
This PR results from the observation that changes to the code base can be made without detecting that the existing functionality was broken.

**Note:**
This PR requires the following PR's to be merged first (rebase this PR than => CI will be green):
- [x] PR #24

### DIFF
--- a/test/include/psen_scan_v2/async_barrier.h
+++ b/test/include/psen_scan_v2/async_barrier.h
@@ -1,0 +1,49 @@
+// Copyright (c) 2020 Pilz GmbH & Co. KG
+//
+// This program is free software: you can redistribute it and/or modify
+// it under the terms of the GNU Lesser General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+//
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU Lesser General Public License for more details.
+//
+// You should have received a copy of the GNU Lesser General Public License
+// along with this program.  If not, see <https://www.gnu.org/licenses/>.
+#ifndef PSEN_SCAN_V2_ASYNC_BARRIER_H
+#define PSEN_SCAN_V2_ASYNC_BARRIER_H
+
+#include <future>
+#include <thread>
+#include <chrono>
+
+namespace psen_scan_v2_test
+{
+class Barrier
+{
+public:
+  void release();
+  template <class Rep, class Period>
+  bool waitTillRelease(const std::chrono::duration<Rep, Period>& timeout) const;
+
+private:
+  std::promise<void> barrier_;
+  const std::future<void> future_{ barrier_.get_future() };
+};
+
+void Barrier::release()
+{
+  barrier_.set_value();
+}
+
+template <class Rep, class Period>
+bool Barrier::waitTillRelease(const std::chrono::duration<Rep, Period>& timeout) const
+{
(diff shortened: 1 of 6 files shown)


martiniil:
We should merge #18 before this one, there could be conflicts.

test/unit_tests/unittest_scanner.cpp
```diff
@@ -39,50 +37,12 @@ static const std::string DEVICE_IP{ "127.0.0.100" };
...
+TEST(ScannerTest, testGetCompleteScan)
 {
-protected:
-  ScannerConfiguration scanner_config_{ HOST_IP,   HOST_UDP_PORT_DATA, HOST_UDP_PORT_CONTROL,
-                                        DEVICE_IP, START_ANGLE,        END_ANGLE };
-};
-
-typedef ScannerT<psen_scan_v2_test::ScannerControllerMock> MockedScanner;
-
-TEST_F(ScannerTest, testStart)
```
martiniil, line 53: Why did you remove these tests?

test/integration_tests/integrationtest_scanner_api.cpp
```diff
@@ -0,0 +1,149 @@
...
+#include "psen_scan_v2/scanner_configuration.h"
+#include "psen_scan_v2/scanner.h"
+#include "psen_scan_v2/start_request.h"
+#include "psen_scan_v2/scanner_reply_msg.h"
+
+namespace psen_scan_v2_test
+{
+static const std::string SCANNER_IP_ADDRESS{ "127.0.0.1" };
+static constexpr unsigned short DATA_PORT_OF_SCANNER_DEVICE{ 2000 };
+static constexpr unsigned short CONTROL_PORT_OF_SCANNER_DEVICE{ 3000 };
```
martiniil, line 38: Only a minor thing. I would rename to `SCANNER_UDP_PORT_DATA` and `SCANNER_UDP_PORT_CONTROL` in accordance to the other constants.

test/include/psen_scan_v2/async_barrier.h
```diff
@@ -0,0 +1,49 @@
...
+#ifndef PSEN_SCAN_V2_ASYNC_BARRIER_H
+#define PSEN_SCAN_V2_ASYNC_BARRIER_H
+
+#include <future>
+#include <thread>
+#include <chrono>
+
+namespace psen_scan_v2_test
+{
+class Barrier
```
martiniil, line 24: This class provides the same mechanism as `AsyncTest` from pilz_testutils. Do you think we could replace `AsyncTest` by this class in the future? Now we have both classes used in this package, which I do not like very much.

martiniil (changes requested):
In general :+1:  for the new tests. However I would keep the unittests of the scanner class. Also see the comments below.

test/unit_tests/unittest_scanner.cpp
hslusarek: Because they are a duplication with regard to the tests executed in the ScannerAPI integrationtest. -> Done.

test/include/psen_scan_v2/async_barrier.h
hslusarek: I agree, however, for me it is just a first step. If the new Barrier class proves itself, I also suggest to replace the old `AsyncTest` class. For now I would keep it that way. Or what do you suggest?

test/integration_tests/integrationtest_scanner_api.cpp
hslusarek: Done.

PR #29, opened 2020-09-25
Feature/new firmware move failure classes

cdoehn:
Reason: The classes list in rosdoc_lite becomes much better overseeable.


include/psen_scan_v2/monitoring_frame_msg.h
```diff
@@ -109,6 +109,15 @@ class MonitoringFrameMsg
 
   uint32_t scan_counter_{ 0 };
   std::vector<double> measures_;
+
+public:
```
hslusarek, line 113: Please, state all public stuff before the private stuff. This way the reader can easily overview the public interface of the class. Furthermore, ensure that first all public methods are stated and then all public members, etc.

I suggest that you state the possible exceptions at the beginning of the class. Please, separate the function declaration and the definition to ease readability.

include/psen_scan_v2/scanner_reply_msg.h
```diff
@@ -94,6 +93,16 @@ class ScannerReplyMsg
...
+public:
+  class CRCMismatch : public std::runtime_error
+  {
+  public:
+    CRCMismatch(const std::string& msg = "CRC did not match") : std::runtime_error(msg)
+    {
+    }
+    virtual ~CRCMismatch() = default;
+  };
 };
```
hslusarek, line 106: Please, state all public stuff before the private stuff. This way the reader can easily overview the public interface of the class. Furthermore, ensure that first all public methods are stated and then all public members, etc.

I suggest that you state the possible exceptions at the beginning of the class. Please, separate the function declaration and the definition to ease readability.

include/psen_scan_v2/udp_client.h
```diff
@@ -135,6 +133,23 @@ class UdpClientImpl
...
+  };
+
+  class OpenConnectionFailure : public std::runtime_error
+  {
+  public:
+    OpenConnectionFailure(const std::string& msg = "Failure while opening connection") : std::runtime_error(msg)
+    {
+    }
+  };
 };
```
hslusarek, line 153: Please, state all public stuff before the private stuff. This way the reader can easily overview the public interface of the class. Furthermore, ensure that first all public methods are stated and then all public members, etc.

I suggest that you state the possible exceptions at the beginning of the class. Please, separate the function declaration and the definition to ease readability.

hslusarek (changes requested):
Just minor things.

PR #30, opened 2020-09-25
Move protocol specific constants to scanner_constants header file

cdoehn:


include/psen_scan_v2/scanner_constants.h
```diff
@@ -16,18 +16,116 @@
...
+static constexpr uint32_t OPCODE{ 0xCA };
+static constexpr uint32_t ONLINE_WORKING_MODE{ 0x00 };
+static constexpr uint32_t GUI_MONITORING_TRANSACTION{ 0x05 };
+
+static constexpr std::size_t OFFSET_OPCODE{ 0x04 };
+static constexpr std::size_t OFFSET_WORKING_MODE{ 0x08 };
+static constexpr std::size_t OFFSET_TRANSACTION_TYPE{ 0x0C };
+static constexpr std::size_t OFFSET_SCANNER_ID{ 0x10 };
+
+class FieldHeader
```
agutenkunst, line 91: why should this be here? This is a class not a constant

hslusarek (changes requested):
Please provide a clear concept on how we should proceed with device constants. For the concept also consider how other software components like exceptions are handled and where they are located. Furthermore, try to use new data types like `enum class` instead of primitive data types.

PR #31, opened 2020-09-28
Feature/new firmware move constants to classes

cdoehn:


agutenkunst:
I think these values from the manual should also appear on `StartRequest::toRawData()` where we currently rely on the type/size of members...

cdoehn:
> I think these values from the manual should also appear on `StartRequest::toRawData()` where we currently rely on the type/size of members...

I think the usage of these data offsets would conflict with the concept of processing the data via a stream.

PR #32, opened 2020-09-28
Remove unused arguments

cdoehn:
keep it simple


martiniil:
I think we should rather make use of the `num_bytes` in the deserialization of the monitoring frame.

Else you should also change `udp_client.h`, where these arguments are passed:
```
using NewDataHandler = std::function<void(const MaxSizeRawData&, const std::size_t&)>;
```

martiniil:
alternative: #34

PR #33, opened 2020-09-29
Better processing of raw data

martiniil:


src/monitoring_frame_msg.cpp
```diff
@@ -71,11 +69,18 @@ MonitoringFrameMsg MonitoringFrameMsg::fromRawData(const MaxSizeRawData& data)
...
+          os << "Length of scan counter field is " << header.length() << ", but should be " << 4 << ".";
+          throw MonitoringFrameFormatErrorScanCounterUnexpectedSize(os.str());
+        }
+        raw_processing::read(is, msg.scan_counter_);
         break;
 
       case AdditionalFieldIds::MEASURES:
-        readMeasures(is, msg.measures_, header.length());
+        raw_processing::readArray<uint16_t, double>(
+            is, msg.measures_, header.length() / 2, [](uint16_t raw_element) { return raw_element / 1000.; });
```
cdoehn, line 83:
```suggestion
            is, msg.measures_, header.length() / sizeof(uint16_t), [](uint16_t raw_element) { return raw_element / 1000.; });
```

cdoehn (changes requested):

src/monitoring_frame_msg.cpp
martiniil: OK?
```
        raw_processing::readArray<uint16_t, double>(is,
                                                    msg.measures_,
                                                    header.length() / NUMBER_OF_BYTES_SINGLE_MEASURE,
                                                    [](uint16_t raw_element) { return raw_element / 1000.; });
```

cdoehn: Even better than my suggestion

cdoehn (approved):

PR #39, opened 2020-10-01
Feature/new firmware improve tenth of degree

hslusarek:
Enforce usage of `TenthOfDegree` class and introduce new `ScanRange` class.

### DIFF
--- a/include/psen_scan_v2/scan_range.h
+++ b/include/psen_scan_v2/scan_range.h
@@ -1,0 +1,91 @@
+// Copyright (c) 2020 Pilz GmbH & Co. KG
+//
+// This program is free software: you can redistribute it and/or modify
+// it under the terms of the GNU Lesser General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+//
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU Lesser General Public License for more details.
+//
+// You should have received a copy of the GNU Lesser General Public License
+// along with this program.  If not, see <https://www.gnu.org/licenses/>.
+
+#ifndef PSEN_SCAN_V2_SCAN_RANGE_H
+#define PSEN_SCAN_V2_SCAN_RANGE_H
+
+#include <stdexcept>
+
+#include "psen_scan_v2/tenth_of_degree.h"
+
+namespace psen_scan_v2
+{
+/**
+ * @brief Higher level data type storing the range in which the scanner takes measurements.
+ */
+template <uint16_t min_allowed_angle, uint16_t max_allowed_angle>
+class ScanRange
+{
+  static_assert(min_allowed_angle < max_allowed_angle, "MIN-angle limit smaller than MAX-angle limit.");
+
+public:
+  ScanRange() = default;
+
+  /**
+   * @brief Constructor.
+   *
+   * @param start_angle Start angle of measurement (scanner-zero = zero on the left).
+   * @param end_angle End angle of measurement.
+   */
+  constexpr ScanRange(const TenthOfDegree& start_angle, const TenthOfDegree& end_angle);
+
+  const TenthOfDegree& getStart() const;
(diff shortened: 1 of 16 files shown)


test/unit_tests/unittest_scan_range.cpp
```diff
@@ -0,0 +1,95 @@
...
+{
+static constexpr TenthOfDegree MIN_ALLOWED_ANGLE{ 7 };
+static constexpr TenthOfDegree MAX_ALLOWED_ANGLE{ 54 };
+
+static constexpr TenthOfDegree TOO_SMALL_SCAN_ANGLE{ MIN_ALLOWED_ANGLE.value() - 1u };
+static constexpr TenthOfDegree TOO_LARGE_SCAN_ANGLE{ MAX_ALLOWED_ANGLE.value() + 1u };
+
+using TestScanRange = ScanRange<MIN_ALLOWED_ANGLE.value(), MAX_ALLOWED_ANGLE.value()>;
+
+class ScanRangeBuilder
```
agutenkunst, line 34: Maybe this is a little bit over the top? How about just using the CTor in the Tests?

test/unit_tests/unittest_scan_range.cpp
```diff
@@ -0,0 +1,95 @@
...
+  TenthOfDegree start_angle_{ MIN_ALLOWED_ANGLE.value() + 1u };
+  TenthOfDegree end_angle_{ MAX_ALLOWED_ANGLE.value() - 1u };
+};
+
+TEST(ScanRangeTest, testCtorCallForCoverage)
+{
+  std::unique_ptr<DefaultScanRange> ptr{ new DefaultScanRange(MIN_ALLOWED_ANGLE, MAX_ALLOWED_ANGLE) };
+}
+
+TEST(ScanRangeTest, testTooSmallStartAngle)
```
agutenkunst, line 64:
```suggestion
TEST(ScanRangeTest, testStartAngleTooSmall)
```

include/psen_scan_v2/scan_range.h
```diff
@@ -0,0 +1,89 @@
...
+#include <stdexcept>
+
+#include "psen_scan_v2/tenth_of_degree.h"
+
+namespace psen_scan_v2
+{
+/**
+ * @brief Higher level data type storing the range in which the scanner takes measurements.
+ */
+template <uint16_t min_allowed_angle, uint16_t max_allowed_angle>
```
agutenkunst, line 28: I am not that comfortable with values as template parameters since for every set of template arguments a class is compiled this does not scale well. Also you run into issues if you want to implement operators etc... see https://godbolt.org/z/1jf5vj

In this case at hand you also don't check for min_allowed_angle <= max_allowed_angle

Maybe ScanRange could have an (optional?) limits parameter of type ScanRange? Or there is LimitedScanRange inheriting from ScanRange?

agutenkunst (changes requested):
Some findings, see inline. Maybe split the ScanRange into a separate PR ?

include/psen_scan_v2/scan_range.h
agutenkunst: This would also not work if the angles would have to be double.

agutenkunst, line 28: Maybe a static_assert if max_allowed_angle > min_allowed_angle?

test/unit_tests/unittest_scan_range.cpp
hslusarek: Done.

test/unit_tests/unittest_scan_range.cpp
hslusarek: Done.

hslusarek:
> Maybe split the ScanRange into a separate PR ?

I decided against it because of the hard coded angle limits in the `ScannerConfiguration` class. The limit issues needs to be addressed if one wants to enforce the usage of the `TenthOfDegree` class and at the same time get all tests too run. It would be unnecessary work overhead to split these limits out, just to integrate them in a separate PR into the new data structure `ScanRange`.

include/psen_scan_v2/scan_range.h
hslusarek: Done.

PR #42, opened 2020-10-06
Use original state machine in unittest_scanner_controller

martiniil:


test/include/psen_scan_v2/mock_udp_client.h
```diff
@@ -18,31 +18,103 @@
...
                     const std::chrono::high_resolution_clock::duration& timeout));
   // "Simulates" function call which uses default values
   MOCK_METHOD0(startAsyncReceiving, void());
   MOCK_METHOD1(write, void(const psen_scan_v2::DynamicSizeRawData& data));
+
+private:
+  void handleNewData(const psen_scan_v2::MaxSizeRawData& received_data, const std::size_t& bytes_received);
+
+private:
+  psen_scan_v2::NewDataHandler data_handler_;
```
agutenkunst, line 71: I think `using psen_scan_v2` is legit in tests / mocks

test/unit_tests/unittest_scanner_controller.cpp
```diff
@@ -65,182 +68,151 @@ class ScannerControllerTest : public ::testing::Test
...
+  const MaxSizeRawData raw_data = convertToMaxSizeRawData(test_data.hex_dump);
+  const auto num_bytes = 2 * test_data.hex_dump.size();
+  const MonitoringFrameMsg frame{ MonitoringFrameMsg::fromRawData(raw_data, num_bytes) };
+  return toLaserScan(frame);
 }
 
-TEST_F(ScannerControllerTest, testStopRequestSending)
+TEST_F(ScannerControllerTest, testStart)
 {
   using ::testing::_;
```
agutenkunst, line 112: Could this be at toplevel?

test/unit_tests/unittest_scanner_controller.cpp
```diff
@@ -65,182 +68,151 @@ class ScannerControllerTest : public ::testing::Test
...
-  using ::testing::Expectation;
+  using ::testing::InSequence;
 
-  StopRequest stop_request;
+  const StartRequest start_request(scanner_config_, 0);
```
agutenkunst, line 115: Maybe us this inplace?

test/unit_tests/unittest_scanner_controller.cpp
```diff
@@ -65,182 +68,151 @@ class ScannerControllerTest : public ::testing::Test
...
+    InSequence seq;
+    EXPECT_CALL(scanner_controller_.control_udp_client_, startAsyncReceiving(_, _, _)).Times(1);
+    EXPECT_CALL(scanner_controller_.data_udp_client_, startAsyncReceiving()).Times(1);
+    EXPECT_CALL(scanner_controller_.control_udp_client_, write(start_request.toRawData())).Times(1);
+  }
 
-  scanner_controller_.sendStopRequest();
+  auto start_future = scanner_controller_.start();
+  sendStartReply();
+  start_future.wait_for(std::chrono::seconds(0));
```
agutenkunst, line 126: ?

test/unit_tests/unittest_scanner_controller.cpp
```diff
@@ -65,182 +68,151 @@ class ScannerControllerTest : public ::testing::Test
...
-  EXPECT_CALL(scanner_controller_.control_udp_client_, write(stop_request.toRawData())).Times(1);
+  {
+    InSequence seq;
+    EXPECT_CALL(scanner_controller_.control_udp_client_, startAsyncReceiving(_, _, _)).Times(1);
+    EXPECT_CALL(scanner_controller_.data_udp_client_, startAsyncReceiving()).Times(1);
+    EXPECT_CALL(scanner_controller_.control_udp_client_, write(start_request.toRawData())).Times(1);
+  }
 
-  scanner_controller_.sendStopRequest();
+  auto start_future = scanner_controller_.start();
```
agutenkunst, line 124:
```suggestion
  scanner_controller_.start();
```
is not possible?

test/unit_tests/unittest_scanner_controller.cpp
```diff
@@ -65,182 +68,151 @@ class ScannerControllerTest : public ::testing::Test
...
+  sendStartReply();
+  start_future.wait_for(std::chrono::seconds(0));
 }
 
-TEST_F(ScannerControllerTest, testHandleStartReplyTimeout)
+TEST_F(ScannerControllerTest, testStartReplyTimeout)
 {
   using ::testing::_;
```
agutenkunst, line 131: See above

test/unit_tests/unittest_scanner_controller.cpp
```diff
@@ -65,182 +68,151 @@ class ScannerControllerTest : public ::testing::Test
...
+LaserScan ScannerControllerTest::testDataToLaserScan(const TestData& test_data)
+{
+  const MaxSizeRawData raw_data = convertToMaxSizeRawData(test_data.hex_dump);
+  const auto num_bytes = 2 * test_data.hex_dump.size();
+  const MonitoringFrameMsg frame{ MonitoringFrameMsg::fromRawData(raw_data, num_bytes) };
+  return toLaserScan(frame);
 }
 
-TEST_F(ScannerControllerTest, testStopRequestSending)
+TEST_F(ScannerControllerTest, testStart)
```
agutenkunst, line 110: Maybe more specific test description

test/unit_tests/unittest_scanner_controller.cpp
```diff
@@ -65,182 +68,151 @@ class ScannerControllerTest : public ::testing::Test
...
 
-  EXPECT_CALL(scanner_controller_.state_machine_, processReplyReceivedEvent(ScannerReplyMsgType::Start)).Times(1);
-  scanner_controller_.handleScannerReply(max_size_data, max_size_data.size());
+  auto stop_future = scanner_controller_.stop();
+  sendStopReply();
+  stop_future.wait_for(std::chrono::seconds(0));
 }
 
-TEST_F(ScannerControllerTest, testHandleScannerReplyTypeStop)
+TEST_F(ScannerControllerTest, testStopReplyTimeout)
```
agutenkunst, line 165: From reading the test I cannot see where a timeout is tested....

agutenkunst: 1. It is expected that receiving is started.... ok.
1. It is expected that a stop request is written... ok.

but nothing about a timeout is expected?

test/unit_tests/unittest_scanner_controller.cpp
```diff
@@ -65,182 +68,151 @@ class ScannerControllerTest : public ::testing::Test
...
+{
+  simulateUdpError("Udp error");  // only for coverage for now
 }
 
 TEST_F(ScannerControllerTest, testConstructorInvalidLaserScanCallback)
 {
-  LaserScanCallback laserscan_callback;
-  typedef ScannerControllerT<psen_scan_v2_test::ControllerStateMachineMock, psen_scan_v2_test::MockUdpClient> SCT;
+  const LaserScanCallback laserscan_callback;
+  typedef ScannerControllerT<ControllerStateMachine, MockUdpClient> SCT;
```
agutenkunst, line 215: I know this was before the same, but y?

test/unit_tests/unittest_scanner_controller.cpp
```diff
@@ -65,182 +68,151 @@ class ScannerControllerTest : public ::testing::Test
...
 }
 
-TEST_F(ScannerControllerTest, testHandleStartReplyTimeout)
+TEST_F(ScannerControllerTest, testStartReplyTimeout)
 {
   using ::testing::_;
-  using ::testing::Expectation;
+  using ::testing::InSequence;
+
+  const StartRequest start_request(scanner_config_, 0);
```
agutenkunst, line 134: Magic number

test/unit_tests/unittest_scanner_controller.cpp
agutenkunst: I thing the issue is that actually timeout is not the tested behavior but the resend is...

test/unit_tests/unittest_scanner_controller.cpp
rfeistenauer: Now expecting correct result

test/unit_tests/unittest_scanner_controller.cpp
rfeistenauer: it is but we though testing for the promise to get fullfilled would be nicer

test/unit_tests/unittest_scanner_controller.cpp
rfeistenauer: Problem is at this point the timeout happens but has no real implementation, we could only test for the log msg and we decided not to do that since the real functionallity is already implemented in another branch, same in testStartReplyTimeout

test/unit_tests/unittest_scanner_controller.cpp
agutenkunst: Yes indeed. Now with `EXPECT_EQ(start_future.wait_for(std::chrono::seconds(0)), std::future_status::ready);` even better!

agutenkunst (approved):
Thank you!

PR #43, opened 2020-10-07
Implement Serialization for MonitoringFrames to be able to generate TestData

cdoehn:


include/psen_scan_v2/raw_processing.h
```diff
@@ -77,6 +77,18 @@ inline void readArray(std::istringstream& is,
   });
 }
 
+template <typename T>
+inline T serialize(std::ostringstream& os)
```
agutenkunst, line 81: Not clear what this function does. Given the name I would assume you can pass it a ostringstream and it returns an arbitrary type...

test/src/monitoring_frame_serialization.cpp
```diff
@@ -0,0 +1,62 @@
...
+  uint32_t scan_counter_header_payload = frame.scan_counter_;
+  raw_processing::write(os, scan_counter_header_id);
+  raw_processing::write(os, scan_counter_header_length);
+  raw_processing::write(os, scan_counter_header_payload);
+
+  FieldHeader::Id measures_header_id = AdditionalFieldIds::MEASURES;
+  FieldHeader::Length measures_header_length = frame.measures_.size() * NUMBER_OF_BYTES_SINGLE_MEASURE + 1;
+  raw_processing::write(os, measures_header_id);
+  raw_processing::write(os, measures_header_length);
+  for (auto& elem : frame.measures_)
```
agutenkunst, line 46: Consider extending `raw_processing::`. See `raw_processing::readArray`

include/psen_scan_v2/raw_processing.h
cdoehn: renamed to toArray

PR #57, opened 2020-10-14
Swap tf to match rotation axis of scanner

ct2034:
Related PR for tutorial https://github.com/PilzDE/pilz_tutorials/pull/23


urdf/psen_scan.urdf.xacro
```diff
@@ -0,0 +1,102 @@
...
+      <origin rpy="${pi} 0 0" xyz="0 ${radius - dist_scanning_axis_rear} ${height / 2 - dist_scanning_plane_top}"/>
+    </joint>
+    
+    <!-- This is the location of the laser scan -->
+    <link name="${prefix}_scan">
+    </link>
+  </xacro:macro>
+
+
+  <xacro:macro name="psen_scan_gazebo_v0" params="name link ros_topic update_rate min_angle max_angle min_range max_range">
```
agutenkunst, line 69: Do we need this?

ct2034: not at the moment. I think later we should definitely also think about how to best help costumers with their gazebo model. And this might just be it. But for now, it is not necessary.

README.md
```diff
@@ -100,6 +100,24 @@ roslaunch psen_scan_v2 psen_scan_v2.launch sensor_ip:=192.168.0.10 host_ip:=192.
 ```
 This example configures the safety laser scanner at 192.168.0.10 to send it´s frames to 192.168.0.20:3050.
 
+## TF Frames
+The location of the TF frames is shown in the image below.
+Note that these names defined by the `prefix` parameter.
+This can be changed with aforementioned launchfile parameters.
+For example when you are running multiple scanners.
```
rfeistenauer, line 107:
```suggestion
> These names are defined by the aforementioned launchfile parameter: `prefix` \
> Changing them is necessary for instance when running multiple scanners.
```

README.md
```diff
@@ -100,6 +100,24 @@ roslaunch psen_scan_v2 psen_scan_v2.launch sensor_ip:=192.168.0.10 host_ip:=192.
...
+<p align="center">
+<img src="img/frames.png" width="800px" alt="PILZ safety laser scanner frames" title="frames">
+</p>
```
rfeistenauer, line 110: I think side and rear view already give all necessary information for this topic. Maybe omit the other two at this point?

README.md
```diff
@@ -81,8 +81,8 @@ UDP Port used to send commands (start/stop) and receive the corresponding replie
 _sensor_ip_ (_string_, default: "192.168.0.10")<br/>
 IP-Address of safety laser scanner.
 
-_frame_id_ (_string_, default: "scanner")<br/>
-Identifier used for transformations within ROS environment.
+_prefix_ (_string_, default: "laser_1")<br/>
```
rfeistenauer, line 84: is this a prefix also used by other vendors or could we name it like psenscan_1 to have it related to the pilz product?

test/unit_tests/unittest_laserscan_ros_conversions.cpp
```diff
@@ -44,34 +47,31 @@ TEST(LaserScanROSConversionsTest, testToLaserScanMsg)
...
   EXPECT_EQ(laserscan_msg.range_max, RANGE_MAX_IN_M);
 
   ASSERT_EQ(laserscan_msg.ranges.size(), measures.size());
 
-  // Check that the ranges in the ROS msg is reversed in comparision to the laserscan and given in meters
-  auto reverse_it_measures = measures.crbegin();
-  auto iter_ranges = laserscan_msg.ranges.cbegin();
-  while (reverse_it_measures != measures.rend() && iter_ranges != laserscan_msg.ranges.end())
+  // Check that the ranges in the ROS msg is the same order as the laserscan and given in meters
+  for (size_t i = 0; i < laserscan_msg.ranges.size(); ++i)
```
rfeistenauer, line 72: sounds like this could be an extra test?

urdf/psen_scan.urdf.xacro
```diff
@@ -0,0 +1,58 @@
+<?xml version="1.0"?>
+<robot xmlns:xacro="http://ros.org/wiki/xacro">
+  <xacro:macro name="psen_scan" params="prefix">
+    <xacro:property name="min_range" value="0.1" /> <!-- TODO correct dimensions -->
```
rfeistenauer, line 4: why do dimensions have to be corrected?

rfeistenauer (changes requested):

README.md
ct2034: I like them for clarity

README.md
ct2034: other vendors use a similar method. I think we have the psenscan branding already in many places.

urdf/psen_scan.urdf.xacro
ct2034: Come to think of it, we don't need them any more ;-)

PR #58, opened 2020-10-15
Introduce new Scanner owning the state machine

hslusarek:
This PR removes the obsolete components `ScannerController` and `ControllerStateMachine`. Their code is integrated into the new `ScannerV2` and `ScannerProtocolDef` implementation classes. 
The removal of the `ScannerController` and `ControllerStateMachine` components removes the need to pass callbacks from the `ScannerController` to the actual state machine (unfortunately some callbacks still need to be passed -> see note below). At the same time it is now possible to directly use the state machine with all their features.

**Note:**
It is still necessary to pass certain things from the `ScannerV2` class to the `ScannerProtocolDef` class. This is due to the fact that the state machine front-end (`msm::front::state_machine_def<...>`) has no access to the state machine back-end (`msm::back::state_machine<...>`) which, in turn, holds important functions like `StateMachine.start()`, `StateMachine.stop()`, `StateMachine.process_event()`, etc.


include/psen_scan_v2/scanner_v2.h
```diff
@@ -0,0 +1,102 @@
...
+  // a "std::unique_ptr" to "msm::front::state_machine_def".
+  StateMachineArgs* createStateMachineArgs();
+
+  template <class T>
+  void triggerEventWithParam(const T& event);
+
+  template <class T>
+  void triggerEvent();
+
+  void scannerStartedCB();
```
agutenkunst, line 68: As far as I understood the code: If `startStartWatchdog` and `stopStartWatchdog` would be part of the state machine and not of `ScannerV2` the and the future would be passed to the SM using the event (while returning the promise to the user) the scanner would no longer need to pass own callback to the SM and the encapsulation would be stronger... But maybe I am missing something...

hslusarek: I think you mean we could pass the `promise` to the SM and the `future` to the user`, right?

Yes, this is definitely a good idea. I give it a try.

test/integration_tests/integrationtest_scanner_api.cpp
```diff
@@ -165,12 +172,57 @@ TEST(ScannerAPITests, testStartFunctionality)
...
+  std::cout << "ScannerAPITests: Send unexpected monitoring frame..." << std::endl;
+  std::vector<double> measures = { 1, 2, 3, 4, 5 };
+  MonitoringFrameMsg monitoring_frame(TenthOfDegree(0), TenthOfDegree(1), 0, measures);
+  scanner_mock.sendMonitoringFrame(monitoring_frame);
+  EXPECT_EQ(start_future.wait_for(WAIT_TIMEOUT), std::future_status::timeout) << "Scanner::start() finished too early";
+
+  std::cout << "ScannerAPITests: Send start reply..." << std::endl;
+  scanner_mock.sendStartReply();
+  EXPECT_EQ(start_future.wait_for(DEFAULT_TIMEOUT), std::future_status::ready) << "Scanner::start() not finished";
+}
```
rfeistenauer, line 218: If I understand that right this test tests:
- No laser scan callback is called on early monitoring frame
- barrier does open on start request.
- start future is set valid right after start
- start future is set valid by sending monitoring frame
- start future is valid after sending start reply.

The name states does not fail, which can mean a lot, but is all that necessary to test if an early monitoring frame does not change the start behavior?
Maybe adding one or two more tests would split this into easier understandable chunks

test/integration_tests/integrationtest_scanner_api.cpp
```diff
@@ -244,6 +324,46 @@ TEST(ScannerAPITests, testReceivingOfMonitoringFrame)
...
 
+TEST(ScannerAPITests, shouldNotCallLaserscanCallbackInCaseOfEmptyMonitoringFrame)
+{
+  StrictMock<ScannerMock> scanner_mock;
+  const ScannerConfiguration config{ createScannerConfig() };
+  UserCallbacks cb;
+  ScannerV2 scanner(config, std::bind(&UserCallbacks::LaserScanCallback, &cb, std::placeholders::_1));
+
+  std::vector<double> measures = { 1, 2, 3, 4, 5 };
+  MonitoringFrameMsg msg(TenthOfDegree(0), TenthOfDegree(1), 0, measures);
```
rfeistenauer, line 335: Why do we have scanner_mock.sendEmptyMonitoringFrame(); but not something like scanner_mock.sendNormalMonitoringFrame()?

rfeistenauer (changes requested):
I like the new structure.

include/psen_scan_v2/scanner_v2.h
hslusarek: The problem is that a `promise` cannot be copied. The `msm::front::state_machine_def<...>` tries to make a copy of the event which it gets via `trigger_event`. If the `promise` is part of the event, the event cannot be copied. Therefore, the approach sounds good but is not realizable. (Of course a hack would be possible but, in my opinion, this is not the way to go here.)

include/psen_scan_v2/scanner_state_machine.hpp
```diff
@@ -0,0 +1,140 @@
...
+  {\
+    PSENSCAN_DEBUG("StateMachine", fmt::format("Exiting state: {}", #state_name));\
+  }
+// clang-format on
+
+// DEFAULT_STATE_IMPL(Idle)
+DEFAULT_STATE_IMPL(WaitForStartReply)
+// DEFAULT_STATE_IMPL(WaitForMonitoringFrame)
+DEFAULT_STATE_IMPL(WaitForStopReply)
+// DEFAULT_STATE_IMPL(Stopped)
```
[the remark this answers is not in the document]
hslusarek: Removed -> Done.

include/psen_scan_v2/scanner_state_machine.hpp
```diff
@@ -0,0 +1,140 @@
...
+{
+  PSENSCAN_DEBUG("StateMachine", fmt::format("Entering state: {}", "WaitForMonitoringFrame"));
+  fsm.args_->scanner_started_cb();
+}
+
+template <class Event, class FSM>
+void ScannerProtocolDef::WaitForMonitoringFrame::on_exit(Event const&, FSM& fsm)
+{
+  PSENSCAN_DEBUG("StateMachine", fmt::format("Exiting state: {}", "WaitForMonitoringFrame"));
+}
```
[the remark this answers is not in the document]
hslusarek: Done.

test/integration_tests/integrationtest_scanner_api.cpp
hslusarek: >[...] but is all that necessary to test if an early monitoring frame does not change the start behavior?

I removed or changed some of the checks. However, some checks simply are necessary because the test is dealing with more than one thread. To get the timing right, one simply needs to check certain conditions.

And yes a monitoring frame should not change the start behavior and this is exactly what this test tries to check. I'm not sure how I should reduce the test complexity. The scenario is a little bit complicated to reproduce but I don't see how I should simplify the test.

test/integration_tests/integrationtest_scanner_api.cpp
hslusarek: Done.

test/integration_tests/integrationtest_scanner_api.cpp
```diff
@@ -165,12 +172,57 @@ TEST(ScannerAPITests, testStartFunctionality)
   EXPECT_EQ(start_future.wait_for(DEFAULT_TIMEOUT), std::future_status::ready) << "Scanner::start() not finished";
 }
 
+TEST(ScannerAPITests, shouldThrowWhenStartIsCalledTwice)
+{
+  NiceMock<ScannerMock> scanner_mock;
+  const ScannerConfiguration config{ createScannerConfig() };
+  UserCallbacks cb;
+  ScannerV2 scanner(config, std::bind(&UserCallbacks::LaserScanCallback, &cb, std::placeholders::_1));
```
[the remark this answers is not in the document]
hslusarek: No, I purposely decided against it. I want to leave room for new tests to emerge. If I define to much in a test fixture, it becomes increasingly more complicated to add new and maybe different API tests. Each test more or less, represents a possible user action and, therefore, I think it is better if all "user actions" (simulated by the test), can directly be seen in the test function itself (see https://stackoverflow.com/a/11837973).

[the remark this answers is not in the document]
hslusarek: Done.

rfeistenauer (approved):

include/psen_scan_v2/scanner_v2.h
agutenkunst:
```
/////////////////////////////////////
// Compile with g++ <file> -lboost_system -std=c++11 -pthread
/////////////////////////////////////

#include <future>
#include <iostream>
#include <string>

// back-end
#include <boost/msm/back/state_machine.hpp>
// front-end
#include <boost/msm/front/state_machine_def.hpp>

namespace msm = boost::msm;
namespace mpl = boost::mpl;

struct sm_ : public msm::front::state_machine_def<sm_> {
  std::promise<std::string> promise_;

  struct eventB {};

  struct states {
    struct A : public msm::front::state<> {
      template <class Event, class FSM> void on_exit(Event const &, FSM &) {
        std::cerr << "Leaving A\n";
      }
    };
    struct B : public msm::front::state<> {
      template <class Event, class FSM> void on_entry(Event const &, FSM &) {
        std::cerr << "Entering B\n";
      }
    };
  };

  void action(eventB const &) {
    try {
      promise_.set_value("new promise value");
    } catch (const std::future_error &e) {
      std::cerr << "Future error" << '\n';
    } catch (const std::exception &e) {
      std::cerr << e.what() << '\n';
    }
  }

  struct transition_table
      : mpl::vector<
            //    Start          Event             Next       Action  Guard
            //  +--------------+-----------------+-----------+------+-------+
            a_row<states::A, eventB, states::B, &sm_::action>
            //  +--------------+-----------------+-----------+------+-------+
            > {};

  typedef states::A initial_state;
};

// Pick a back-end
typedef msm::back::state_machine<sm_> sm;

int main() {
  sm state_maschine;

  std::promise<std::string> promise;
  std::future<std::string> future = promise.get_future();
  state_maschine.promise_ = std::move(promise);

  state_maschine.process_event(sm_::eventB());
  std::cerr << "Value is \"" << future.get() << "\"\n";
}
```

hslusarek: Yes, we could move the promise into the the state machine front-end. But what happens if the `Scanner::start()` function  is called again? Then we need to reset the promise like it is done in the `ScannerController`. In my opinion, this is not really a good alternative.

agutenkunst (approved):
Checked with real system :heavy_check_mark:

PR #59, opened 2020-10-16
Adds scanner started msg

rfeistenauer:
Changes a debug lvl msg to info to notify the user of a successful start of the scanner.

This is in particular useful when the start was not successful right away.


agutenkunst (approved):
Checked on real system.

PR #63, opened 2020-10-20
Fix fmt headers

agutenkunst:


src/monitoring_frame_msg.cpp
```diff
@@ -21,6 +21,9 @@
 #include <string>
 #include <vector>
 
+#include <fmt/core.h>
+#include <fmt/format.h>
```
martiniil, line 25: Add `<fmt/ostream.h>?

martiniil (changes requested):

src/monitoring_frame_msg.cpp
agutenkunst: I don't thing it is needed here. It should only be requested were a custom type is use by `fmt::format("{}", custom_type_var)` . Including fmt/ostream.h should only be needed where this is used.

martiniil: I think you have a custom type (`std::vector<MonitoringFrameDiagnosticMessage>`) here:
https://github.com/PilzDE/psen_scan_v2/blob/3dc8df0a0ece02a89a625bbde61bcdc82d7d5cf5/src/monitoring_frame_msg.cpp#L76

agutenkunst: Resolved.

martiniil (approved):

PR #64, opened 2020-10-20
Introduce namespaces for MonitoringFrame

cdoehn:
Introduce namespaces for monitoring_frame and monitoring_frame::diagnostic. Also some namespaces for other parts of monitoring frame: additional_field and diagnostic:raw_message;


src/diagnostics.cpp
```diff
@@ -19,7 +19,11 @@ using namespace psen_scan_v2;
 
 namespace psen_scan_v2
 {
-std::ostream& operator<<(std::ostream& os, const MonitoringFrameDiagnosticMessage& msg)
+namespace monitoring_frame
+{
+namespace diagnostic
+{
+std::ostream& operator<<(std::ostream& os, const Message& msg)
```
agutenkunst, line 26:
```suggestion
std::ostream& operator<<(std::ostream& os, const psen_scan_v2::monitoring_frame::diagnostic::Message& msg)
```

include/psen_scan_v2/monitoring_frame_msg.h
```diff
@@ -84,16 +86,17 @@ class MonitoringFrameMsg
...
   std::vector<double> measures_;
   std::vector<double> intensities_;
-  std::vector<MonitoringFrameDiagnosticMessage> diagnostic_messages_;
+  std::vector<monitoring_frame::diagnostic::Message> diagnostic_messages_;
   bool diagnostic_data_enabled_{ false };
 
 public:
-  friend DynamicSizeRawData serialize(const MonitoringFrameMsg& frame);
-  friend MonitoringFrameMsg deserializeMonitoringFrame(const MaxSizeRawData& data, const std::size_t& num_bytes);
+  friend DynamicSizeRawData serialize(const Message& frame);
```
hslusarek, line 93: The code, like here, is inconsistent. Sometimes `Message` and sometimes `monitoring_frame::Message` is used. I suggest to use only one version, preferably, with regard to readability, the following version: 
```suggestion
  friend DynamicSizeRawData serialize(const monitoring_frame::Message& frame);
```

hslusarek:
Add a description to your PR because you are only introducing namespaces for the MonitoringFrame or am I wrong?

include/psen_scan_v2/diagnostics.h
```diff
@@ -26,14 +26,18 @@
...
+namespace raw_message
+{
+static constexpr uint32_t LENGTH_FOR_ONE_DEVICE_IN_BYTES{ 9 };
+static constexpr uint32_t UNUSED_OFFSET_IN_BYTES{ 4 };
+static constexpr uint32_t LENGTH_IN_BYTES{ UNUSED_OFFSET_IN_BYTES +
+                                           LENGTH_FOR_ONE_DEVICE_IN_BYTES * VALID_SCANNER_IDS.size() };
+using Type = std::array<uint8_t, diagnostic::raw_message::LENGTH_IN_BYTES>;
+}  // namespace raw_message
 
 enum class DiagnosticCode
```
hslusarek, line 42: Above it is defined: `raw_message::Type` but here it is defined: `diagnostic::DiagnosticCode`. This is inconsistent. I suggest:
```suggestion
enum class ErrorType
```

include/psen_scan_v2/diagnostics.h
```diff
@@ -137,14 +141,15 @@ class ErrorLocation
...
 public:
-  constexpr MonitoringFrameDiagnosticMessage(const ScannerId& id, const ErrorLocation& location);
+  constexpr Message(const ScannerId& id, const ErrorLocation& location);
 
-  constexpr bool operator==(const MonitoringFrameDiagnosticMessage& rhs) const;
+  constexpr bool operator==(const Message& rhs) const;
 
-  friend RawDiagnosticMsg serializeDiagnosticMessages(const std::vector<MonitoringFrameDiagnosticMessage>& messages);
+  friend diagnostic::raw_message::Type
+  psen_scan_v2::monitoring_frame::diagnostic::serializeMessages(const std::vector<diagnostic::Message>& messages);
```
hslusarek, line 152: No need to state the obvious. The code should state only the namespace part increasing readability.
```suggestion
  diagnostic::serialize(const std::vector<diagnostic::Message>& messages);
```

include/psen_scan_v2/diagnostics.h
```diff
@@ -166,13 +171,12 @@ class MonitoringFrameDiagnosticMessage
...
 
-constexpr inline MonitoringFrameDiagnosticMessage::MonitoringFrameDiagnosticMessage(const ScannerId& id,
-                                                                                    const ErrorLocation& location)
+constexpr inline Message::Message(const ScannerId& id, const ErrorLocation& location)
   : id_(id), error_location_(location)
 {
 }
 
-constexpr inline bool MonitoringFrameDiagnosticMessage::operator==(const MonitoringFrameDiagnosticMessage& rhs) const
+constexpr inline bool Message::operator==(const Message& rhs) const
```
hslusarek, line 179: Code inconsistent.
```suggestion
constexpr inline bool diagnostic::Message::operator==(const diagnostic::Message& rhs) const
```

include/psen_scan_v2/diagnostics.h
```diff
@@ -186,8 +190,10 @@ inline bool isAmbiguous(const DiagnosticCode& code)
   return ambiguous_diagnostic_codes.find(code) != ambiguous_diagnostic_codes.end();
 }
 
-std::ostream& operator<<(std::ostream& os, const MonitoringFrameDiagnosticMessage& msg);
+std::ostream& operator<<(std::ostream& os, const Message& msg);
```
hslusarek, line 193: Code inconsistent.
```suggestion
std::ostream& operator<<(std::ostream& os, const diagnostic::Message& msg);
```

include/psen_scan_v2/monitoring_frame_deserialization.h
```diff
@@ -24,34 +24,17 @@
...
-
-  static std::string idToString(Id id);
-
-private:
-  Id id_;
-  Length length_;
-};
 
-class MonitoringFrameFixedFields
+class FixedFields
```
hslusarek, line 37: I know where the idea comes from but to stay consistent with the additional fields, I suggest to rename it to:
```suggestion
class Header
```
Because if each additional field has a header describing whats coming next, so does the MonitoringFrame itself.
Using this idea, we write for example:
```
using namespace psen_scan_v2;
const monitoring_frame::Header monitoring_frame_header;
const monitoring_frame::additional_field::Header additional_field_header;
```
This is more consistent in my opinion. What do you think?

include/psen_scan_v2/monitoring_frame_deserialization.h
```diff
@@ -88,90 +71,116 @@ class MonitoringFrameFixedFields
...
-std::vector<MonitoringFrameDiagnosticMessage> deserializeDiagnosticMessages(std::istringstream& is);
+Header read(std::istringstream& is, const std::size_t& max_num_bytes);
+}  // namespace additional_field
 
-class MonitoringFrameFormatError : public std::runtime_error
+Message deserialize(const MaxSizeRawData& data, const std::size_t& num_bytes);
+FixedFields readFixedFields(std::istringstream& is);
+namespace diagnostic
+{
+std::vector<diagnostic::Message> deserializeDiagnosticMessages(std::istringstream& is);
```
hslusarek, line 112:
```suggestion
std::vector<diagnostic::Message> deserializeMessages(std::istringstream& is);
```
Because then we can write:
```
diagnostic::deserializeMessages(...);
```
instead of:
```
diagnostic::deserializeDiagnosticMessages(...);
```

hslusarek (changes requested):
I guess a lot of the comments are very subjective, therefore, you do not necessarily need to implement them. However, I think a good discussion about these points is important before we continue.

include/psen_scan_v2/diagnostics.h
cdoehn: I prefer changing the name of diagnostic::raw_message::Type to diagnostic::raw_message::Field

include/psen_scan_v2/diagnostics.h
cdoehn: I get a compiler error message when I do this:

error: extra qualification ‘psen_scan_v2::monitoring_frame::diagnostic::Message::’ on member ‘operator==’ [-fpermissive]

include/psen_scan_v2/monitoring_frame_deserialization.h
cdoehn: At the beginning I called it also header with exact the same idea...

@agutenkunst @martiniil complainted about that.  

To be consistent with the protocol description FixedFields would be better, because MonitoringFrame has no header.

agutenkunst: Yes because there is no such thing as a Header mentioned in the protocol manual. Only Headers for the Additional Fields. Therefore we concluded that FixedFields is the better naming to avoid confusion.

include/psen_scan_v2/diagnostics.h
hslusarek: If in your opinion is the better name then use this name. My point is that the `Diagnostics` should not be included in the name because it is already included in the namespace name.

hslusarek (approved):

PR #66, opened 2020-10-21
Improve robustness of ScannerAPI tests

hslusarek:
This PR tries to make the ScannerAPI tests more robust. Therefore,
- the ports are changed for every test,
- the Timeout test `testStartReplyTimeout()` is modified,
- the test `shouldThrowWhenStartIsCalledTwice()` is temporarily removed because there seems to be an error in the production code which needs to be fixed.

### DIFF
--- a/include/psen_scan_v2/scanner_v2.h
+++ b/include/psen_scan_v2/scanner_v2.h
@@ -47,7 +47,10 @@ static constexpr std::chrono::milliseconds REPLY_TIMEOUT{ 1000 };
 class ScannerV2 : public IScanner
 {
 public:
-  ScannerV2(const ScannerConfiguration& scanner_config, const LaserScanCallback& laser_scan_cb);
+  ScannerV2(const ScannerConfiguration& scanner_config,
+            const LaserScanCallback& laser_scan_cb,
+            const unsigned short data_port_scanner = DATA_PORT_OF_SCANNER_DEVICE,
+            const unsigned short control_port_scanner = CONTROL_PORT_OF_SCANNER_DEVICE);
   ~ScannerV2();
 
 public:
@@ -57,7 +60,8 @@ public:
 private:
   // Raw pointer used here because "msm::back::state_machine" cannot properly pass
   // a "std::unique_ptr" to "msm::front::state_machine_def".
-  StateMachineArgs* createStateMachineArgs();
+  StateMachineArgs* createStateMachineArgs(const unsigned short& data_port_scanner,
+                                           const unsigned short& control_port_scanner);
 
   template <class T>
   void triggerEventWithParam(const T& event);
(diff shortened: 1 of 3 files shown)


include/psen_scan_v2/scanner_v2.h
```diff
@@ -47,7 +47,10 @@ static constexpr std::chrono::milliseconds REPLY_TIMEOUT{ 1000 };
 class ScannerV2 : public IScanner
 {
 public:
-  ScannerV2(const ScannerConfiguration& scanner_config, const LaserScanCallback& laser_scan_cb);
+  ScannerV2(const ScannerConfiguration& scanner_config,
+            const LaserScanCallback& laser_scan_cb,
+            const unsigned short data_port_scanner = DATA_PORT_OF_SCANNER_DEVICE,
```
agutenkunst, line 52: Could/Should these be part of the ScannerConfiguration?

hslusarek: I'm actually not sure. If so, the `ScannerConfiguration` should have default values for it because it is probably only rarely changing.

test/integration_tests/integrationtest_scanner_api.cpp
```diff
@@ -69,20 +64,77 @@ using namespace ::testing;
...
+    control_port_scanner = (control_port_scanner + 1) % MAX_CONTROL_PORT_HOST;
+    if (control_port_scanner == 0)
+    {
+      control_port_scanner = MIN_CONTROL_PORT_HOST;
+    }
+
+    data_port_scanner = (data_port_scanner + 1) % MAX_CONTROL_PORT_HOST;
+    if (data_port_scanner == 0)
+    {
+      data_port_scanner = MIN_CONTROL_PORT_HOST;
```
rfeistenauer: shouldn't this be MAX/MIN_DATA_PORT_SCANNER?

test/integration_tests/integrationtest_scanner_api.cpp
```diff
@@ -69,20 +64,77 @@ using namespace ::testing;
...
-  return ScannerConfiguration(
-      HOST_IP_ADDRESS, HOST_UDP_PORT_DATA, HOST_UDP_PORT_CONTROL, SCANNER_IP_ADDRESS, SCAN_RANGE, DIAGNOSTICS_ENABLED);
-}
+  PortHolder& operator++()
+  {
+    data_port_host = (data_port_host + 1) % MAX_DATA_PORT_HOST;
+    if (data_port_host == 0)
+    {
+      data_port_host = MIN_DATA_PORT_HOST;
+    }
```
rfeistenauer, line 80: Could this be simplified as:
```suggestion
    data_port_host = std::max((data_port_host + 1) % MAX_DATA_PORT_HOST, MIN_DATA_PORT_HOST);
```

rfeistenauer: same for the next 3 I gues

rfeistenauer (commented):
Just a few things for now

test/integration_tests/integrationtest_scanner_api.cpp
hslusarek: Already fixed

test/integration_tests/integrationtest_scanner_api.cpp
hslusarek: Already fixed.

PR #68, opened 2020-10-21
Remove ALL_CAPS from enum class

agutenkunst:
See http://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Renum-caps


cdoehn (commented):
Are this enums not something like constants? Then they should be capital:

http://wiki.ros.org/CppStyleGuide#Constants

I propose not to mix this styles. Let us decide for one rule

agutenkunst:
Style was mixed before. Have a close look at all enums before this PR. Also check the reasoning at http://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Renum-caps

agutenkunst:
Ok seems that the ROS-Style wants https://clang.llvm.org/extra/clang-tidy/checks/readability-identifier-naming.html#cmdoption-arg-enumconstantcase to be UPPER_CASE see https://github.com/ros-planning/moveit/blob/master/.clang-tidy

Not sure what https://clang.llvm.org/extra/clang-tidy/checks/readability-identifier-naming.html#cmdoption-arg-scopedenumconstantcase does.

Also clang-tidy only detects the violations in cpp files. Thus lower_case is allowed in *.h files. 

How to proceed?

martiniil:
> Style was mixed before. Have a close look at all enums before this PR. Also check the reasoning at http://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Renum-caps

This suggests that ALL_CAPS is reserved for macros only. I think we don't need to do that.

agutenkunst:
> > Style was mixed before. Have a close look at all enums before this PR. Also check the reasoning at http://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Renum-caps
> 
> This suggests that ALL_CAPS is reserved for macros only. I think we don't need to do that.

So change .clang-tidy?

martiniil:
> > > Style was mixed before. Have a close look at all enums before this PR. Also check the reasoning at http://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Renum-caps
> > 
> > 
> > This suggests that ALL_CAPS is reserved for macros only. I think we don't need to do that.
> 
> So change .clang-tidy?

No, I suggest we keep it like that.

cdoehn:
I changed .clang-tidy to case style lower_case...
On my opinion it doesn't matter, which case style but it has to be consistent

martiniil:
> I changed .clang-tidy to case style lower_case...
> On my opinion it doesn't matter, which case style but it has to be consistent

Ok for me

PR #72, opened 2020-10-22
Add intensity feature

hslusarek:
This PR tells the scanner to also measure the intensities.


hslusarek:
Due to merge error closed and continued in PR #75.

PR #74, opened 2020-10-23
Notify user about start retry

agutenkunst:
This was broken by 43a790317cd7589576e33356d25b17ff67272edb

### DIFF
--- a/include/psen_scan_v2/scanner_state_machine.h
+++ b/include/psen_scan_v2/scanner_state_machine.h
@@ -118,6 +118,8 @@ public:  // States
 public:  // Action methods
   template <class T>
   void sendStartRequest(const T& event);
+  template <class T>
+  void handleStartRequestTimeout(const T& event);
   void sendStopRequest(const scanner_events::StopRequest& event);
   void handleMonitoringFrame(const scanner_events::RawMonitoringFrameReceived& event);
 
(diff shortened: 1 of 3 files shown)


include/psen_scan_v2/scanner_state_machine.hpp
```diff
@@ -81,6 +81,16 @@ inline void ScannerProtocolDef::sendStartRequest(const T& event)
   args_->control_client_->write(StartRequest(args_->config_, DEFAULT_SEQ_NUMBER).serialize());
 }
 
+template <class T>
+inline void ScannerProtocolDef::handleStartRequestTimeout(const T& event)
+{
+  PSENSCAN_DEBUG("StateMachine", "Action: handleStartRequestTimeout");
+  PSENSCAN_ERROR("StateMachine",
+                 "Timeout while waiting for the scanner to start! Retrying... "
+                 "(Please check the ethernet connection or contact PILZ support if the error persists.)");
```
hslusarek, line 90: If this was missing in the StateMachine refactoring and if it is still important, it is probably best for the future to add a test which checks for this messages.

include/psen_scan_v2/scanner_state_machine.h
```diff
@@ -118,6 +118,8 @@ class ScannerProtocolDef : public msm::front::state_machine_def<ScannerProtocolD
 public:  // Action methods
   template <class T>
   void sendStartRequest(const T& event);
+  template <class T>
+  void handleStartRequestTimeout(const T& event);
```
hslusarek, line 122: Why is this action a template function? It does not have to be because this action is only called with one event (`StartTimeout`).

hslusarek (changes requested):
Minor things.

include/psen_scan_v2/scanner_state_machine.h
hslusarek: Fixed -> Done.

include/psen_scan_v2/scanner_state_machine.hpp
hslusarek: Done.

hslusarek (approved):

PR #76, opened 2020-10-26
Fix/multi start

hslusarek:
This PR tries to remedy the issues which occur if `Scanner::start()` is called twice.

### DIFF
--- a/include/psen_scan_v2/scanner_v2.h
+++ b/include/psen_scan_v2/scanner_v2.h
@@ -21,6 +21,9 @@
 #include <future>
 #include <chrono>
 #include <functional>
+#include <atomic>
+
+#include <boost/optional.hpp>
 
 #include "psen_scan_v2/scanner_interface.h"
 #include "psen_scan_v2/scanner_events.h"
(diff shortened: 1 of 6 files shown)


src/scanner_v2.cpp
```diff
@@ -111,10 +112,10 @@ void ScannerV2::scannerStartedCB()
   PSENSCAN_INFO("ScannerController", "Scanner started successfully.");
   stopStartWatchdog();
   PSENSCAN_DEBUG("Scanner", "Inform user that scanner start is finsihed.");
-  scanner_has_started_.set_value();
+  std::for_each(scanner_has_started_.begin(), scanner_has_started_.end(), [](auto& promise) { promise.set_value(); });
 
   // Reinitialize
-  scanner_has_started_ = std::promise<void>();
+  scanner_has_started_.clear();
```
hslusarek, line 118: Due to the fact that the `ScannerV2::start()` and the `ScannerV2::scannerStartedCB` functions are called from at least two different threads, namely the user-main thread and the `io_service` thread form the `UdpClient`, the promise vector needs to be protected against concurrent access via mutex or something.
If the promise vector is protected via mutex, one actually can skip the vector usage and continue using only one promise protected via mutex (or realiased via atomic).

include/psen_scan_v2/scanner_v2.h
```diff
@@ -76,7 +76,7 @@ class ScannerV2 : public IScanner
   void stopStartWatchdog();
 
 private:
-  std::promise<void> scanner_has_started_;
+  std::vector<std::promise<void>> scanner_has_started_;
```
hslusarek, line 79: Only `ScannerV2::start()` is considered, however, the problem is the same for the `ScannerV2::stop()`.

src/scanner_v2.cpp
hslusarek: Note: `std::promise` cannot be used in combination with `std::atomic` because:
```
error: static assertion failed: std::atomic requires a trivially copyable type
       static_assert(__is_trivially_copyable(_Tp),
```

hslusarek:
Closed -> Continued in PR #78.

PR #77, opened 2020-10-26
Enable conversion of negative radian angle values

martiniil:


agutenkunst:
I am not sure what this does. Could you provide some description in the PR? Sounds like negative rad end up being converted to positive tenthDegree...

agutenkunst (changes requested):
Please explain reasoning

agutenkunst:
> * The class `TenthOfDegree` represents degree values in the range [0, 360] multiplied by ten.
> * The static method `TenthOfDegree::fromRad()` can also convert negative radian values to a `TenthOfDegree` object.

I think the confusing thing would be that `TenthOfDegree::fromRad(-1).toRad() != -1` maybe having a signed and an unsigned version ob this could help?

agutenkunst:
Or we use `int32_t` internally and provide checks where it MUST be positive (only StartRequest comes to my mind...)

martiniil:
> Or we use `int32_t` internally and provide checks where it MUST be positive (only StartRequest comes to my mind...)

Would it be acceptable to cast a positive `int32_t` to `uint16_t`?

agutenkunst:
> > Or we use `int32_t` internally and provide checks where it MUST be positive (only StartRequest comes to my mind...)
> 
> Would it be acceptable to cast a positive `int32_t` to `uint16_t`?

Only if it fits...

include/psen_scan_v2/angle_conversions.h
```diff
@@ -30,29 +33,42 @@ inline static constexpr double degreeToRadian(const double& angle_in_degree)
...
   return degreeToTenthDegree(radianToDegree(angle_in_rad));
 }
 
-inline static double constexpr tenthDegreeToRad(const uint16_t& angle_in_tenth_degree)
+inline static double constexpr tenthDegreeToRad(const int32_t& angle_in_tenth_degree)
 {
   return degreeToRadian(static_cast<double>(angle_in_tenth_degree) / 10.);
 }
 
+inline static uint16_t tenthDegreeToPositiveTenthDegree(const int32_t& angle_in_tenth_degree)
```
SansoneG, line 59: I feel like `tenthDegreeToPositiveTenthDegree` and `degreeToTenthDegree` should return the same type. Any reason against this?

include/psen_scan_v2/tenth_of_degree.h
```diff
@@ -49,7 +54,7 @@ class TenthOfDegree
     return *this;
   }
 
-  constexpr TenthOfDegree& operator*(const unsigned int& rhs)
+  constexpr TenthOfDegree& operator*(const int& rhs)
```
SansoneG, line 57: What happens then this overflows?

include/psen_scan_v2/tenth_of_degree.h
```diff
@@ -28,12 +28,17 @@ class TenthOfDegree
...
+  }
+
 public:
-  explicit constexpr TenthOfDegree(const uint16_t& tenth_of_degree) : tenth_of_degree_(tenth_of_degree)
+  explicit constexpr TenthOfDegree(const int32_t& tenth_of_degree) : tenth_of_degree_(tenth_of_degree)
   {
   }
 
-  constexpr uint16_t value() const
+  constexpr int32_t value() const
```
SansoneG, line 41: I feel like `value()` and `toPositiveValue` should return the same type

test/unit_tests/unittest_start_request.cpp
```diff
@@ -98,9 +98,11 @@ TEST_F(StartRequestTest, constructorTest)
...
   EXPECT_TRUE(DecodingEquals<uint8_t>(data, static_cast<size_t>(Offset::DIAGNOSTICS_ENABLED), 0));
 
-  EXPECT_TRUE(DecodingEquals(data, static_cast<size_t>(Offset::MASTER_START_ANGLE), scan_range.getStart().value()));
-  EXPECT_TRUE(DecodingEquals(data, static_cast<size_t>(Offset::MASTER_END_ANGLE), scan_range.getEnd().value()));
-  EXPECT_TRUE(DecodingEquals(data, static_cast<size_t>(Offset::MASTER_ANGLE_RESOLUTION), degreeToTenthDegree(0.2)));
+  EXPECT_TRUE(
+      DecodingEquals(data, static_cast<size_t>(Offset::MASTER_START_ANGLE), (uint16_t)scan_range.getStart().value()));
+  EXPECT_TRUE(
+      DecodingEquals(data, static_cast<size_t>(Offset::MASTER_END_ANGLE), (uint16_t)scan_range.getEnd().value()));
+  EXPECT_TRUE(DecodingEquals(data, static_cast<size_t>(Offset::MASTER_ANGLE_RESOLUTION), (uint16_t)2));
```
SansoneG, line 105:
```suggestion
      DecodingEquals(data, static_cast<size_t>(Offset::MASTER_START_ANGLE), static_cast<uint16_t>(scan_range.getStart().value())));
  EXPECT_TRUE(
      DecodingEquals(data, static_cast<size_t>(Offset::MASTER_END_ANGLE), static_cast<uint16_t>(scan_range.getEnd().value())));
  EXPECT_TRUE(DecodingEquals(data, static_cast<size_t>(Offset::MASTER_ANGLE_RESOLUTION), static_cast<uint16_t>(2)));
```

include/psen_scan_v2/angle_conversions.h
```diff
@@ -30,29 +33,42 @@ inline static constexpr double degreeToRadian(const double& angle_in_degree)
...
+{
+  if (angle_in_tenth_degree < std::numeric_limits<uint16_t>::min() ||
```
SansoneG, line 61: Why does a function that has `toPositive` in it reject negative numbers? This function basically only checks whether a given angle is positive and could therefore also return a `bool` at that point.

include/psen_scan_v2/angle_conversions.h
martiniil: Actually the method name was misleading. I renamed it to `tenthDegreeToUnsignedTenthDegree()`.

include/psen_scan_v2/tenth_of_degree.h
martiniil: I renamed `toPositiveValue` to `toUnsigned`. I hope now the intention is clear.

include/psen_scan_v2/angle_conversions.h
martiniil: Again, the method-name was misleading. I hope now the intention is clear.

include/psen_scan_v2/tenth_of_degree.h
martiniil: I don't know. Do you suggest that we detect overflow of integer arithmetic?

martiniil:
Needs more revision. Close in favor of #84 for the sake of clarity.

PR #78, opened 2020-10-27
Fix multi Scanner::start() issues

hslusarek:
This PR tries to remedy the issues which appear when `ScannerV2::start()` is called more than once.

### DIFF
--- a/include/psen_scan_v2/scanner_v2.h
+++ b/include/psen_scan_v2/scanner_v2.h
@@ -21,6 +21,8 @@
 #include <future>
 #include <functional>
 
+#include <boost/optional.hpp>
+
 #include "psen_scan_v2/scanner_interface.h"
 #include "psen_scan_v2/scanner_events.h"
 #include "psen_scan_v2/scanner_state_machine.h"
@@ -81,17 +83,26 @@ private:
   };
 
 private:
-  std::promise<void> scanner_has_started_;
-  std::promise<void> scanner_has_stopped_;
+  using OptionalPromise = boost::optional<std::promise<void>>;
 
-  std::mutex sm_mutex_;
+private:
+  OptionalPromise scanner_has_started_{ boost::none };
+  OptionalPromise scanner_has_stopped_{ boost::none };
+
+  //! @brief This Mutex protects ALL members of the Scanner against concurrent access.
+  //! So far there exist at least the following threads, potentially causing concurrent access to the members:
+  //! - user-main-thread
+  //! - io_service thread of UDPClient
+  //! - watchdog threads
+  std::mutex member_mutex_;
+
   std::unique_ptr<ScannerStateMachine> sm_;
 };
 
 template <class T>
 void ScannerV2::triggerEventWithParam(const T& event)
 {
-  const std::lock_guard<std::mutex> lock(sm_mutex_);
+  const std::lock_guard<std::mutex> lock(member_mutex_);
   sm_->process_event(event);
 }
 
--- a/src/scanner_v2.cpp
+++ b/src/scanner_v2.cpp
@@ -83,71 +83,72 @@ ScannerV2::ScannerV2(const ScannerConfiguration& scanner_config,
   : IScanner(scanner_config, laser_scan_cb)
   , sm_(new ScannerStateMachine(createStateMachineArgs(data_port_scanner, control_port_scanner)))
 {
-  const std::lock_guard<std::mutex> lock(sm_mutex_);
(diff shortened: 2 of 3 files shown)


include/psen_scan_v2/scanner_v2.h
```diff
@@ -72,26 +74,31 @@ class ScannerV2 : public IScanner
...
-  std::promise<void> scanner_has_stopped_;
+  OptionalPromise scanner_has_started_{ boost::none };
+  OptionalPromise scanner_has_stopped_;
+
+  //! @brief This Mutex protects ALL members of the Scanner against concurrent access.
+  //! So far there exist at least the following threads, potentially causing concurrent access to the members:
+  //! - user-main-thread
+  //! - io_service thread of UDPClient
+  //! - watchdog threads
+  std::recursive_mutex member_mutex_;
```
martiniil, line 89: I don't see the need for a recursive mutex. So I would use `std::mutex`.

martiniil (changes requested):
Looks good. Is it already tested with the hardware setup?
See also the inline comments.

include/psen_scan_v2/scanner_v2.h
hslusarek: No, its needed because of the `triggerEventWithParam()` function. The only other option would be to use an internal version of the `triggerEventWithParam()` function which not automatically locks the mutex.

include/psen_scan_v2/scanner_v2.h
```diff
@@ -72,26 +74,31 @@ class ScannerV2 : public IScanner
...
-  void startStartWatchdog();
-  void stopStartWatchdog();
+private:
+  using OptionalPromise = boost::optional<std::promise<void>>;
 
 private:
-  std::promise<void> scanner_has_started_;
-  std::promise<void> scanner_has_stopped_;
+  OptionalPromise scanner_has_started_{ boost::none };
+  OptionalPromise scanner_has_stopped_;
```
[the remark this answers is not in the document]
hslusarek: Fixed -> Done.

include/psen_scan_v2/scanner_v2.h
hslusarek: I tried it with `std::mutex` and using `sm_->process_event(event);` directly  (in other words without taking the mutex). It does not work, probably because a call to  `sm_->process_event(event);` might cause callbacks to be called which, in turn, also want lock the mutex. In this case the program, consequently, ends in a deadlock.

agutenkunst: Guess by tracking the locks the need for a recursive mutex should be easy to verify

agutenkunst:
Checked on real system :heavy_check_mark:

martiniil (approved):

martiniil:
The scanner api test is unstable:
```
[ RUN      ] ScannerAPITests.shouldReturnInvalidFutureWhenStartIsCalledSecondTime

Host ports:

- data port = 45002

- control port = 57002

Scanner ports:

- data port = 3002

- control port = 7002

[ INFO] [1603975091.467062407]: Start scanner called.

[ INFO] [1603975091.467212924]: Start scanner called.

ScannerMock: Send start reply...

[ INFO] [1603975091.467460264]: Scanner started successfully.

Segmentation fault (core dumped)
```

hslusarek:
> The scanner api test is unstable.

I investigated the test in more detail. It seems there are at least two bugs which actually have not so much to do with this PR. The PR only seems to make the bugs appear more often. Anyhow, this is what I found out:
- The UDPClient from time to time hangs in the [`startAsyncReceive()`](https://github.com/PilzDE/psen_scan_v2/tree/melodic-devel/include/psen_scan_v2/udp_client.h#L254) method.
- The Segmentation fault seems to appear during the destruction of the `ScannerV2` class. However, I have reason to believe that the actual problem is the [`UdpClient()::close`](https://github.com/PilzDE/psen_scan_v2/tree/melodic-devel/include/psen_scan_v2/udp_client.h#L180) method. The shutdown of the socket and the io_service does not seem to be correct.

hslusarek:
Ok, I further investigated the problem and found out that a `Barrier` on the Testmethod stack is called during destruction of the Testmethod. The following happens:
While the Test/Testmethod stack is destroyed the UDPMockServer received a `StartRequest` message and calls it corresponding mock handler. This leads to a call to the `Barrier` which is, however, by then already destroyed or, currently, in destruction. Consequently a segmentation fault appears. I solved the issue by deleting the `Barrier` because the `Barrier` is not needed for the test anyway.

PR #84, opened 2020-10-30
Use signed integer type for representing angles in tenth degree

martiniil:
This essentially replaces `uint16_t` by `int16_t` such that the class `TenthOfDegree` and the angle conversions support negative values.

The protocol does not specify if the 16bit angles are signed or unsigned. But, concerning the scan range, it has to be in [0, 2750], such that it makes no difference if we use signed or unsigned type in (de)serialization.


agutenkunst (changes requested):
Looks good!

Some remarks:
- Somewhere during the deserialization there should be a check if the conversion is safe
- I suppose the existing tests could/be extended

agutenkunst (approved):

PR #89, opened 2020-11-03
Add method to get state name

hslusarek:
Adds a functions allowing to get the name of a state of the state machine by ID.

PR #90, opened 2020-11-03
Improve sporadically hanging empty msg test in scanner api test

hslusarek:
Due to the fact that the empty monitoring frame test sometimes hangs, I redesigned the test a little bit. I'm not sure why the original test stopped sporadically. The improved version however, works even if stress tested and is much more faster.


hslusarek:
When PR #89 is merged this test needs to be adapted, otherwise, it will fail.

martiniil (approved):
I approve the code change. Though I have one question: Which real-world scenario does this test represent? At first I thought in this test a monitoring frame with empty measures-field is sent, which should also not call the laserscan callback.

hslusarek:
> I approve the code change. Though I have one question: Which real-world scenario does this test represent? At first I thought in this test a monitoring frame with empty measures-field is sent, which should also not call the laserscan callback.

Exaclty, it is tested that an empty LaserScan does not call the laserscan callback.

PR #97, opened 2020-11-10
Fix friend declaration

agutenkunst:
Tbh I am not sure why this is not detected earlier but it shows up if you run against noetic using
```
rosrun industrial_ci run_ci ROS_DISTRO="noetic" UPSTREAM_WORKSPACE='github:PilzDE/pilz_robots#fix/missing_test_dep_pilz_utils -pilz_robots/pilz_control -pilz_robots/prbt_hardware_support -pilz_robots/pilz_status_indicator_rqt -pilz_robots/prbt_moveit_config -pilz_robots/prbt_support -pilz_robots/pilz_robots -pilz_robots/prbt_gazebo -pilz_robots/prbt_ikfast_manipulator_plugin'
```

Errors are
```
In file included from /root/target_ws/src/psen_scan_v2/src/diagnostics.cpp:18:
/root/target_ws/src/psen_scan_v2/include/psen_scan_v2/diagnostics.h:150:111: error: ‘psen_scan_v2::monitoring_frame::diagnostic::raw_message::Field psen_scan_v2::monitoring_frame::diagnostic::serialize(const std::vector<psen_scan_v2::monitoring_frame::diagnostic::Message>&)’ should have been declared inside ‘psen_scan_v2::monitoring_frame::diagnostic’
  150 |   friend diagnostic::raw_message::Field diagnostic::serialize(const std::vector<diagnostic::Message>& messages);
```
and
```
In file included from /root/target_ws/src/psen_scan_v2/include/psen_scan_v2/monitoring_frame_deserialization.h:22,
                 from /root/target_ws/src/psen_scan_v2/src/monitoring_frame_deserialization.cpp:22:
/root/target_ws/src/psen_scan_v2/include/psen_scan_v2/monitoring_frame_msg.h:95:108: error: ‘psen_scan_v2::monitoring_frame::Message psen_scan_v2::monitoring_frame::deserialize(const MaxSizeRawData&, const size_t&)’ should have been declared inside ‘psen_scan_v2::monitoring_frame’
   95 |                                                                                const std::size_t& num_bytes);
```

PR #98, opened 2020-11-10
Fix friend declaration

martiniil:
This is a preparation for a noetic release. Since the default distro for noetic is Ubuntu 20 which comes with gcc 9 this error gets detected. (Was wrong even before but gcc7 does not detect it)

Replaces #97


agutenkunst (approved):
Thank you. I added a bit of explantation.

PR #99, opened 2020-11-11
Adapt to new logger API

agutenkunst:
This is caused by https://github.com/PilzDE/pilz_robots/commit/e9a0b37f7104f645e3cba6f2636bfb375d6d7e37 but was possibly not detected due to using the system installed headers which is fixed in https://github.com/PilzDE/pilz_robots/pull/466

A double release of pilz_robots and psen_scan_v2 is required to fix the situation.


hslusarek (approved):
Looks good. Can be merged in `main` as soon as `pilz_robots` was released.

PR #101, opened 2020-11-17
Move Scanner port to ScannerConfiguration

hslusarek:
This PR moves the scanner ports to the `ScannerConfiguration` class. Because the number of  parameters of the `ScannerConfiguration` reached a critical "mass", the `ScannerConfigurationBuilder` is introduced.


include/psen_scan_v2/scanner_config_builder.h
```diff
@@ -0,0 +1,121 @@
+#ifndef PSEN_SCAN_V2_SCANNER_CONFIG_BUILDER_H
```
martiniil, line 1: Please add the copyright header

martiniil (approved):

PR #102, opened 2020-11-18
Make invalid ScanRange creation explicit

hslusarek:
This PR improves the code readability by removing implicit creation of invalid `ScanRange` objects.


martiniil (commented):
I agree that this improves the code readability. But:

I am not really glad with the fact that a default constructed `LaserScanSettings` contains an invalid scan range.
If this problem would be resolved, there would be no need for improving the code readability in this way.

martiniil (approved):

PR #103, opened 2020-11-18
Add noetic to GitHub actions jobs

hslusarek:
This PR adds the GitHub action support for ROS `noetic`.

ToDo
- [x] uncomment unittest_monitoring_frame_msg and fix usage of fmt
- [x] Is there a better way to discriminate between different googletest versions than using the ROS_VERSION_MINOR macro? (especially for pure googletests)
- [x] Update readme: main branch targets melodic & noetic


.github/workflows/industrial_ci_action.yml
```diff
@@ -9,48 +9,41 @@ env:
...
-    continue-on-error: true
-
-  build_test_testing:
-    name: "Build + Test with Testing Repo of melodic (http://packages.ros.org/ros-testing/ubuntu)"
-    runs-on: ubuntu-18.04
+          - {ROS_DISTRO: melodic, ROS_REPO: testing}
+          - {ROS_DISTRO: melodic, ROS_REPO: main}
+          - {ROS_DISTRO: noetic,  ROS_REPO: testing}
+          - {ROS_DISTRO: noetic,  ROS_REPO: main}
+    runs-on: ubuntu-latest
```
SansoneG, line 20: According to https://docs.github.com/en/free-pro-team@latest/actions/reference/workflow-syntax-for-github-actions#github-hosted-runners `ubuntu-latest` currently refers to `ubuntu-18.04`.
Maybe it will work regardless, because the industrial ci uses another docker container itself?

.github/workflows/industrial_ci_action.yml
```diff
@@ -9,48 +9,41 @@ env:
...
         env:
-          ROS_REPO: main
-          ROS_DISTRO: melodic
-    continue-on-error: true
-
-  build_test_testing:
-    name: "Build + Test with Testing Repo of melodic (http://packages.ros.org/ros-testing/ubuntu)"
-    runs-on: ubuntu-18.04
+          - {ROS_DISTRO: melodic, ROS_REPO: testing}
+          - {ROS_DISTRO: melodic, ROS_REPO: main}
```
SansoneG, line 17: How exactly are we going to handle cases, where `testing` runs but `main` doesn't?
See https://docs.github.com/en/free-pro-team@latest/actions/reference/workflow-syntax-for-github-actions#jobsjob_idcontinue-on-error

.github/workflows/industrial_ci_action.yml
hslusarek: Done.

.github/workflows/industrial_ci_action.yml
hslusarek: Added `continue-on-error` which should resolve the issue. -> Done.

.github/workflows/ci_action_melodic.yml
```diff
@@ -0,0 +1,30 @@
+name: CI-Melodic
```
agutenkunst, line 1: Why do we need two seperate workflow files? What is the advantage?

.github/workflows/ci_action_noetic.yml
```diff
@@ -1,4 +1,4 @@
-name: CI
+name: CI-Noetic
```
agutenkunst, line 1: The diff seems to indicate this was was actually `git mv` from the original?

.github/workflows/ci_action_noetic.yml
```diff
@@ -7,50 +7,47 @@ env:
...
+    runs-on: ubuntu-20.04
     steps:
       - uses: actions/checkout@v1
       - uses: 'ros-industrial/industrial_ci@master'
         env:
           ROS_REPO: testing
-          ROS_DISTRO: melodic
           CLANG_FORMAT_CHECK: file
-          CLANG_FORMAT_VERSION: 3.9
+          CLANG_FORMAT_VERSION: 10
```
agutenkunst, line 41: What is the reason behind this?

test/unit_tests/unittest_raw_processing.cpp
```diff
@@ -30,7 +30,7 @@ class RawProcessingTest : public ::testing::Test
 };
 
 using MyTypes = testing::Types<uint8_t, uint16_t, uint32_t, uint64_t>;
-#if (ROS_VERSION_MINOR >= 13)
+#if (ROS_VERSION_MINOR > 14)
```
agutenkunst, line 35: Maybe add a comment that this is due to the different Gtest-Versions

martiniil: Done

.github/workflows/ci_action_noetic.yml
martiniil: I don't know, why it was `3.9` before. Do you know the reason?

Anyways, now the version corresponds to the default on ubuntu focal.

.github/workflows/ci_action_melodic.yml
martiniil: I think it is just a matter of taste. Correct, @SansoneG ?
I like it, because it is less confusing.

.github/workflows/ci_action_noetic.yml
martiniil: I don't know. Maybe this file simply has more similarities with the original than `ci_action_melodic.yml`?

martiniil:
The test-build with clang-tidy enabled fails on noetic. I can reproduce this in a docker container. It says that it cannot match a template member function when trying to format a `psen_scan_v2::monitoring_frame::Message`.

I don't understand that this only happens when clang-tidy is enabled. It seems the following check fails but should be true:
```
is_constructible<fmt::v6::internal::fallback_formatter<psen_scan_v2::monitoring_frame::Message, char, void>>::value
```

I attach the full error log. 
[catkin_Log.txt](https://github.com/PilzDE/psen_scan_v2/files/5643519/catkin_Log.txt)

.github/workflows/ci_action_melodic.yml
SansoneG: The primary reason was to be able to generate CI-Badges for melodic and noetic separately.
Otherwise it is a matter of preference as @martiniil mentioned.

agutenkunst:
Please add the information to the Readme that `main` targets `melodic` and `noetic`.

rfeistenauer (approved):

PR #107, opened 2020-12-01
Enable separate building of cpp-lib.

SansoneG:
To try it yourself:
https://github.com/PilzDE/psen_scan_v2/blob/feature/separate_cpp_lib/standalone/README.md

Open points:
- [x] ~How do we handle getting the fmt sources?~ Directly from installed libfmt package.
- [x] Clean up CMakeLists.txt

Things to add, update or check by the maintainers before this PR can be merged.

- [x]  Public api function documentation
- [x]  Architecture documentation reflects actual code structure
- [x]  Tutorials
- [x]  Overview on ROS wiki
- [x]  Package Readme (example pilz_robots)
- [x]  Good commit messages (some tips)
- [x]  CHANGELOG.rst updated
- [x]  Copyright headers
- [x]  Examples

Review Checklist

- [x]  Soft- and hardware architecture (diagrams and description)
- [x]  Test review (test plan and individual test cases)
- [x]  Documentation describes purpose of file(s) and responsibilities
- [x]  Code (coding rules, style guide)

Release planning (please answer)
- [ ]  When is the new feature released?
- [ ]  Which dependent packages do have to be released simultaneously?

Testing:
- [x] Test standalone lib with real hardware (bionic)
- [x] Test standalone lib with real hardware (focal)

PR #108, opened 2020-12-01
Directly use fmt lib instead of rosfmt

martiniil:
Needed for #107

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -41,8 +41,6 @@ catkin_package(
     sensor_msgs
   DEPENDS
     console_bridge
-  LIBRARIES
-    fmt::fmt
 )
 
 
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -186,6 +186,7 @@ if(CATKIN_ENABLE_TESTING)
   )
   target_link_libraries(unittest_raw_processing
     ${catkin_LIBRARIES}
+    fmt::fmt
   )
 
   catkin_add_gtest(unittest_start_request
@@ -230,6 +231,7 @@ if(CATKIN_ENABLE_TESTING)
   )
   target_link_libraries(unittest_udp_client
     ${catkin_LIBRARIES}
+    fmt::fmt
   )
 
   catkin_add_gtest(unittest_tenth_degree_conversion
@@ -255,6 +257,7 @@ if(CATKIN_ENABLE_TESTING)
   )
   target_link_libraries(unittest_laserscan_conversions
     ${catkin_LIBRARIES}
+    fmt::fmt
   )
 
   catkin_add_gtest(unittest_laserscan_ros_conversions
@@ -265,6 +268,7 @@ if(CATKIN_ENABLE_TESTING)
   )
   target_link_libraries(unittest_laserscan_ros_conversions
     ${catkin_LIBRARIES}
+    fmt::fmt
   )
 
   catkin_add_gtest(unittest_scanner_reply_msg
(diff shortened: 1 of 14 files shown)


martiniil:
Seems not the best idea ... https://github.com/fmtlib/fmt/commit/24924128e32fba8011037e21454426458f99d13b

agutenkunst:
> Seems not the best idea ... [fmtlib/fmt@2492412](https://github.com/fmtlib/fmt/commit/24924128e32fba8011037e21454426458f99d13b)

I don't get it...

agutenkunst:
> Only on Ubuntu 20.04?

Please try to be more descriptive even with Draft PRs and explain the reasoning behind the PR

martiniil:
> > Seems not the best idea ... [fmtlib/fmt@2492412](https://github.com/fmtlib/fmt/commit/24924128e32fba8011037e21454426458f99d13b)
> 
> I don't get it...

Thanks for asking. We have to look also how this would perform for noetic, since libfmt-dev version 6 has some linker errors with gcc >= 8.

CMakeLists.txt
```diff
@@ -41,6 +41,8 @@ catkin_package(
     sensor_msgs
   DEPENDS
     console_bridge
+  LIBRARIES
+    fmt::fmt
```
SansoneG, line 45: We do not need to export the fmt library, do we? (see http://wiki.ros.org/catkin/CMakeLists.txt)

martiniil: I think since our library links against fmt, others who link against us need this. tbh I am not 100% sure.

martiniil: Or this is only relevant if we exported our includes. Then we would not need to list any dependencies in `catkin_package()` at all.

SansoneG: > I think since our library links against fmt, others who link against us need this. tbh I am not 100% sure.

If we want to do this we would need to do `DEPENDS fmt` inside of `catkin_package()`, but it might not be needed, since we already link it against our executable.
(see https://answers.ros.org/question/58498/what-is-the-purpose-of-catkin_depends/?answer=58593#post-id-58593 and the comment of mcamurri)

martiniil: I think you are right. https://github.com/PilzDE/psen_scan_v2/pull/110

test/unit_tests/unittest_monitoring_frame_msg.cpp
```diff
@@ -214,7 +214,7 @@ TEST(MonitoringFrameMsgPrintTest, testPrintMessageSuccess)
   monitoring_frame::Message msg(TenthOfDegree(1234), TenthOfDegree(56), 78, { 45, 44, 43, 42 });
   EXPECT_EQ(fmt::format("{}", msg),
             "monitoring_frame::Message(fromTheta = 123.4 deg, resolution = 5.6 deg, scanCounter = 78, "
-            "measures = {45.0, 44.0, 43.0, 42.0}, intensities = {}, diagnostics = {})");
+            "measures = {45, 44, 43, 42}, intensities = {}, diagnostics = {})");
```
martiniil, line 217: This is due to different formatting of floating point numbers in libfmt version 4. At this point we probably have to distinguish between the melodic and noetic build later.

include/psen_scan_v2/format_range.h
```diff
@@ -0,0 +1,44 @@
...
+#define PSEN_SCAN_V2_FORMAT_RANGE_H
+
+#include <algorithm>
+#include <iterator>
+
+#include <sstream>
+#include <string>
+
+#include <fmt/format.h>
+
```
SansoneG, line 26:
```suggestion

/*
 * The resulting string will look like the following:
 *   "{Item1, Item2, [...], ItemLast}"
 * Examples:
 *   "{1, 2, 3}"
 *   "{}" (Empty Range)
 */

```

SansoneG (approved):

PR #114, opened 2020-12-08
Using latest clang format

ct2034:
It currently only has one effect in our code, but that looks better.
And I am generally in favour of using latest versions of things.


martiniil:
No need for this, is already done in #103

ct2034:
okay, then we should merge that one ;-)

martiniil:
Merged here: 687a1c1

PR #115, opened 2020-12-08
Rename *.hpp file

### LINKED ISSUE #106: why do we have hpp and h files
Why have `psen_scan_v2/include/psen_scan_v2/scanner_state_machine.hpp` and `psen_scan_v2/include/psen_scan_v2/scanner_state_machine.h`. This is very confusing and subjectively somewhat unorthodox. Can we please get rid of one of the two?
And why is the whole state machine implemented ina header anyways? What is the benfit?

martiniil:
## Description

These changes Fix #106


jschleicher (approved):
Does that change anything regarding the doxygen / rosdoc_lite process? There was currently a warning about the .hpp not beeing found?

martiniil:
> 
> 
> Does that change anything regarding the doxygen / rosdoc_lite process? There was currently a warning about the .hpp not beeing found?

Unfortunately not. The warning appears because it is included from within a namespace...

PR #139, opened 2021-01-26
Fix undefined behaviour from unittest_raw_processing

agutenkunst:
## Description

The data read from the ostringstream must be copied byte-wise not in chunks of `sizeof(TypeParam)`.
This show up in failing tests on https://build.ros.org/job/Ndev_db__psen_scan_v2__debian_buster_amd64/15/ and
also explains why it worked with `TypeParam` == uint8_t


martiniil (approved):
👍 Yep, this makes sense.

PR #140, opened 2021-01-27
fix some headers

jschleicher:
reported by include-what-you-use and forward-declare ScannerConfiguration

The full log is attached:
[iwyu.out.txt](https://github.com/PilzDE/psen_scan_v2/files/5881341/iwyu.out.txt)

I've cleaned the header files as suggested. If anyone wants to additionally clean up the cpp files, see the output above.

## Description

These changes clean up included headers.

### Review Checklist
* [x] Code (coding rules, style guide)


agutenkunst:
Thank @jschleicher could you open a separate PR for including this in the CI? Just so that it stays this way...

agutenkunst (approved):
thank you!

PR #162, opened 2021-02-26
fixup CI for standalone branch

jschleicher:
fixup the CI. If putting the `include(ctest)` into the conditional block, we have to also enable this block for windows.


martiniil (approved):
Great!

PR #202, opened 2021-06-07
Fix ubuntu standalone ci

agutenkunst:
## Description

Fixes issues with the standalone build. Could be related to some cmake update I guess...

See for example:
https://github.com/PilzDE/psen_scan_v2/runs/2762716878?check_suite_focus=true

<skip>

### DIFF
--- a/.github/workflows/ci_action_standalone.yml
+++ b/.github/workflows/ci_action_standalone.yml
@@ -3,7 +3,7 @@ name: CI-Standalone
 on: [push, pull_request]
 
 env:
-  BUILD_TYPE: ""
+  BUILD_TYPE: "Release"
 
 jobs:
   build:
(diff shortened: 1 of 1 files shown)


.github/workflows/ci_action_standalone.yml
```diff
@@ -25,12 +25,12 @@ jobs:
...
       shell: bash
       working-directory: ${{runner.workspace}}/build
-      run: cmake $GITHUB_WORKSPACE/standalone -DCMAKE_BUILD_TYPE=$BUILD_TYPE
+      run: cmake $GITHUB_WORKSPACE/standalone -DCMAKE_BUILD_TYPE=$BUILD_TYPE -DBUILD_TESTING=ON
 
     - name: Build
       working-directory: ${{runner.workspace}}/build
       shell: bash
-      run: cmake --build . --config $BUILD_TYPE -DBUILD_TESTING=ON
+      run: cmake --build . --config $BUILD_TYPE
```
martiniil, line 33:
```suggestion
      run: cmake --build .
```
see https://stackoverflow.com/questions/24460486/cmake-build-type-is-not-being-used-in-cmakelists-txt/24470998#24470998 or https://cmake.org/cmake/help/latest/manual/cmake.1.html#build-a-project

martiniil (changes requested):

martiniil:
Should we also recommend a *Release*-build in the README?

martiniil:
It would have been nice to include a link to a failing ci job, however I could not find any.

agutenkunst:
Link to failing job added into description.

martiniil (approved):

PR #211, opened 2021-06-18
Introduce bringup.launch for starting only the scanner node

martiniil:
## Description

These changes allow the system integrator to bringup the scanner from his custom launch file.
The tutorials are changed accordingly: https://github.com/PilzDE/pilz_tutorials/pull/29

To avoid confusion of the user the prefix of the node name is changed to match the prefix of the LaserScan-topic. In detail setting the prefix implies the following namings:
- node: `<prefix>_node`
- topic: `<prefix>_node/scan`
- tf frames: `<prefix>`, `<prefix>_mount_link`, `<prefix>_scan`

### Requirements
- [x] #216 merged

- [x] ~~Public api function documentation~~
- [x] ~~Architecture documentation reflects actual code structure~~
- [x] [Tutorials](https://wiki.ros.org/pilz_robots/Tutorials/)
- [x] ~~Overview on [ROS wiki](https://wiki.ros.org/pilz_robots)~~
- [x] Package Readme ([example pilz_robots](https://github.com/PilzDE/pilz_robots/blob/melodic-devel/README.md))
- [x] CHANGELOG.rst updated
- [x] Copyright headers
- [x] ~~Examples~~

### Review Checklist
- [x] ~~Soft- and hardware architecture (diagrams and description)~~

- [x] When is the new feature released?
asap
- [x] Which dependent packages do have to be released simultaneously?
None

### Hardware tests
- [x] Perform hardware tests

### DIFF
--- a/launch/bringup.launch
+++ b/launch/bringup.launch
@@ -20,7 +20,8 @@ along with this program.  If not, see <https://www.gnu.org/licenses/>.
   <!-- IP-Address of Safety laser scanner -->
   <arg name="sensor_ip" default="192.168.0.10" />
 
-  <!-- Prefix i. e. name of the scanner (required to run multiple scanners) -->
+  <!-- Prefix i. e. name of the scanner (required to run multiple scanners)
+       By convention this should also be used as prefix for the urdf description. -->
   <arg name="prefix" default="laser_1" />
 
   <!-- Start angle of measurement in radian -->
@@ -48,7 +49,7 @@ along with this program.  If not, see <https://www.gnu.org/licenses/>.
   <!-- Set the following to true in order to publish scan data as soon as a UDP packet is ready, instead of waiting for a full scan -->
   <arg name="fragmented_scans" default="false" />
 
-  <node name="laser_scanner" type="psen_scan_v2_node" pkg="psen_scan_v2" output="screen" required="true">
+  <node name="$(arg prefix)_node" type="psen_scan_v2_node" pkg="psen_scan_v2" output="screen" required="true">
     <param name="sensor_ip" value="$(arg sensor_ip)" />
     <param name="prefix" value="$(arg prefix)" />
     <param name="angle_start" value="$(arg angle_start)" />
--- a/launch/bringup.launch
+++ b/launch/bringup.launch
@@ -15,7 +15,7 @@ You should have received a copy of the GNU Lesser General Public License
 along with this program.  If not, see <https://www.gnu.org/licenses/>.
 -->
 
-<!-- Includes starting the scanner driver together with a robot model and preconfigured rviz. -->
+<!-- Use this launch-file for starting the scanner driver exclusively. -->
 <launch>
   <!-- IP-Address of Safety laser scanner -->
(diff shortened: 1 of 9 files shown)


launch/bringup.launch
```diff
@@ -0,0 +1,64 @@
...
+  <!-- UDP Port on which monitoring frames (scans) should be received -->
+  <arg name="host_udp_port_data" default="55115" />
+
+  <!-- UDP Port used to send commands (start/stop) and receive the corresponding replies -->
+  <arg name="host_udp_port_control" default="55116" />
+
+  <!-- Set the following to true in order to publish scan data as soon as a UDP packet is ready, instead of waiting for a full scan -->
+  <arg name="fragmented_scans" default="false" />
+
+  <node name="$(arg prefix)_scanner" type="psen_scan_v2_node" pkg="psen_scan_v2" output="screen" required="true">
```
jschleicher, line 51: Suggestion: Could the suffix be dropped?
```suggestion
  <node name="$(arg prefix)" type="psen_scan_v2_node" pkg="psen_scan_v2" output="screen" required="true">
```

martiniil:
Check: Can we set the params from a .yaml-file after including bringup.launch?

launch/bringup.launch
agutenkunst: I think having an arg `prefix` which actually ends up beeing the `node_name` is strange. The word is **pre**fix.

martiniil: Ok, but `laser_1_scanner` is also strange. How about using the prefix as namespace, like for the prbt? Then we could name it
`laser_1/scanner` or
`laser_1/driver` or
`scanner_1/driver`
What do you think?

martiniil:
> 
> 
> Check: Can we set the params from a .yaml-file after including bringup.launch?

The answer is yes.

launch/bringup.launch
rfeistenauer: Well you could choose another prefix if 'laser_1_scanner' sounds strange. Maybe 'first_laser_scanner'? 
Imho using a namespace is ok as well. Would still prefer the word 'scanner' then.

ct2034: how about the suffix being `_node`.
then it would be `laser_1_node`

ct2034: and for the tf frames it is still a prefix. @agutenkunst

martiniil: Using Christians proposal now. `<prefix>_node` = `laser_1_node`

jschleicher: Like this you still have different names for tf an node namespace?! I think, the idea was to have unique names??

agutenkunst: @martiniil maybe you could state in the description how setting the prefix influences:

- topicname
- tf name
- node name

Its hard to keep track of all the implications...

martiniil: see description ...

agutenkunst: This sounds good to me! Would merge

agutenkunst (approved):

PR #227, opened 2021-08-24
Enable output on failure for standalone ci jobs

martiniil:
Here is an example output with a test failure:

https://github.com/martiniil/psen_scan_v2/runs/3408852133


agutenkunst (approved):
thank you!

PR #231, opened 2021-09-03
Automate timestamp hardware-test

martiniil:
## Description

Transforms the acceptance test "standalone/test/acceptance_tests/acceptancetest_timestamps into an hardware test that can be run automated by the HW CI. The new test also includes the test cases from the former `hwtest_timestamp.py`.

Further explanation for the test procedure can be found [here](https://github.com/PilzDE/psen_scan_v2/blob/feature/automate_timestamp_hwtest/test/hw_tests/hwtest_readme.md#hardware-test-hwtest_timestamp_standalone)

ToDo:
- [x] Test with industrial_ci
- [x] Document usage
- [x] Update github_runner alias
- [x] is `hwtest_timestamp.py` now redundant?

- [x] ~CHANGELOG.rst updated~
- [x] Copyright headers
- [x] ~Examples~

### Review Checklist

### Hardware tests
- [x] Perform hardware tests


test/hw_tests/hwtest_timestamp_standalone.py
```diff
@@ -0,0 +1,153 @@
...
+
+    def test_timestamp(self):
+        self.wait_for_msgs()
+        print("Done Captureing\n")
+        lines = read_udp_pakets(self.udp_pakets)
+        stamps = read_timestamp_output(self.timestamp_output)
+        print(f"have {len(lines)} packages and {len(stamps)} callback calls")
+
+        pairs = pair_udp_data_and_stamps(lines, stamps)
+        for last_sr, sr in zip(pairs, pairs[1:]):
```
martiniil, line 140: If you use it like that, `pairs` should be ordered with respect to the scan counters. (Which it probably is, but we need to be shure)

test/hw_tests/hwtest_timestamp_standalone.py
```diff
@@ -0,0 +1,153 @@
...
+        process.kill()
+    return full_output.strip()
+
+
+def pair_udp_data_and_stamps(udp_data: Sequence[UdpData], stamps: Sequence[TimeStamp]):
+    pairs = []
+    for t in stamps:
+        first = None
+        last = None
+        for u in udp_data:
```
martiniil, line 96:
```suggestion
        for u in filter(t.scan_counter == u.scan_counter, udp_data):
```
should work? But first see the comment below about ordering `pairs`.

martiniil: or maybe we could even generate a list of all frames and use `min`/`max` operators.

test/hw_tests/hwtest_timestamp_standalone.cpp
```diff
@@ -0,0 +1,145 @@
...
+  void SetUp() override
+  {
+    const char* udp_data_filename{ std::getenv(UDP_DATA_FILENAME_ENV_VAR.c_str()) };
+    ASSERT_NE(udp_data_filename, nullptr)
+        << "Searching environment variable " << UDP_DATA_FILENAME_ENV_VAR << "failed.";
+    udp_data_filename_ = udp_data_filename;
+
+    buildScannerConfig();
+    setLogLevel(CONSOLE_BRIDGE_LOG_INFO);
+    ASSERT_GE(testSize(), MINIMUM_TEST_SIZE) << "Assembling test data failed.";
```
rfeistenauer, line 58: Do I understand this correct, that the testSize() in the setUp actually calls testData() which in turn calls assemble() which again runs the scanner for a few seconds?
So testSize() has the side effect of running the scanner for several seconds?

test/hw_tests/hwtest_timestamp_standalone.cpp
```diff
@@ -0,0 +1,145 @@
...
+  }
+
+protected:
+  const TestData& testData() const
+  {
+    static const std::unique_ptr<TestData> test_data_ptr{ TestDataAssembler::assemble(
+        *scanner_config_, SCANNER_RUN_DURATION_S, udp_data_filename_, HOST_UDP_DATA_PORT) };
+    return *test_data_ptr;
```
rfeistenauer, line 66: The static is there to prevent having to gather the data again? Why don't you use a member instead?
And I think you might be creating copies here.

test/include/psen_scan_v2/test_data_assembler.h
```diff
@@ -0,0 +1,55 @@
...
+#include "psen_scan_v2_standalone/scanner_interface.h"
+
+#include "psen_scan_v2/test_data.h"
+#include "psen_scan_v2/udp_data_reader.h"
+
+namespace psen_scan_v2_test
+{
+using namespace psen_scan_v2_standalone;
+
+class TestDataAssembler
```
rfeistenauer, line 33: Does this really have to be a class if it only contains static methods?

test/src/udp_data_reader.cpp
```diff
@@ -0,0 +1,98 @@
...
+
+#include <fmt/format.h>
+
+#include "psen_scan_v2_standalone/util/logging.h"
+
+#include "psen_scan_v2/udp_data_reader.h"
+
+namespace psen_scan_v2_test
+{
+void UdpDataReader::read(const std::string& filename, const uint16_t port, UdpData& udp_data)
```
rfeistenauer, line 32: this method could be cleaner with some extracted methods

test/hw_tests/hwtest_timestamp_standalone.cpp
```diff
@@ -0,0 +1,155 @@
...
+  }
+
+protected:
+  void assembleTestDataByRunningTheScanner()
+  {
+    // We use a static variable in order to avoid multiple setups with data assembly.
+    // If not for https://github.com/google/googletest/issues/247 this could be done via SetUpTestSuite().
+    static const std::shared_ptr<test_data::TestData> test_data_ptr{ test_data::assemble(
+        *scanner_config_, SCANNER_RUN_DURATION_S, udp_data_filename_, HOST_UDP_DATA_PORT) };
+    test_data_ptr_ = test_data_ptr;
```
martiniil, line 70:
```suggestion
    static const std::shared_ptr<test_data::TestData> TEST_DATA_PTR{ test_data::assemble(
        *scanner_config_, SCANNER_RUN_DURATION_S, udp_data_filename_, HOST_UDP_DATA_PORT) };
    test_data_ptr_ = TEST_DATA_PTR;
```

rfeistenauer (approved):
Thank you!

PR #234, opened 2021-09-20
Refactor/tests

rfeistenauer:
## Description

These changes improve the readibility of the initigration tests.

- [x] ~CHANGELOG.rst updated~
- [x] ~Copyright headers~
- [x] ~Examples~

### Review Checklist
- [x] ~Documentation describes purpose of file(s) and responsibilities~

### Hardware tests
- [x] ~Perform hardware tests~

### DIFF
--- a/standalone/test/integration_tests/api/integrationtest_scanner_api.cpp
+++ b/standalone/test/integration_tests/api/integrationtest_scanner_api.cpp
@@ -65,6 +65,10 @@ public:
   MOCK_METHOD1(LaserScanCallback, void(const LaserScan&));
 };
 
+#define EXPECT_FUTURE_IS_READY(val1) EXPECT_EQ(val1.wait_for(DEFAULT_TIMEOUT), std::future_status::ready)
+
+#define EXPECT_FUTURE_NOT_READY(val1) EXPECT_EQ(val1.wait_for(DEFAULT_TIMEOUT), std::future_status::timeout)
+
 class ScannerAPITests : public testing::Test
 {
 protected:
@@ -79,8 +83,6 @@ protected:
   std::unique_ptr<util::Barrier> prepareStartRequestBarrier(const ScannerConfiguration& config);
   std::unique_ptr<util::Barrier>
   prepareMonitoringFrameBarrier(const std::vector<data_conversion_layer::monitoring_frame::Message>& msgs);
-  void expectNotDoneYet(const std::string& method_name, const std::future<void>& future);
-  void expectIsDone(const std::string& method_name, const std::future<void>& future);
   void sendMonitoringFrames(const std::vector<data_conversion_layer::monitoring_frame::Message>& msgs);
 
 protected:
@@ -199,17 +201,7 @@ void ScannerAPITests::startScanner()
     nice_scanner_mock_->startListeningForControlMsg();
   }
   const auto start_future = scanner_->start();
-  expectIsDone("Scanner::start()", start_future);
-}
-
-void ScannerAPITests::expectNotDoneYet(const std::string& method_name, const std::future<void>& future)
-{
-  EXPECT_EQ(future.wait_for(FUTURE_WAIT_TIMEOUT), std::future_status::timeout) << method_name << " finished to early";
-}
-
-void ScannerAPITests::expectIsDone(const std::string& method_name, const std::future<void>& future)
-{
-  EXPECT_EQ(future.wait_for(DEFAULT_TIMEOUT), std::future_status::ready) << method_name << " not finished";
+  EXPECT_FUTURE_IS_READY(start_future) << "Scanner::start() not finished";
 }
 
 void ScannerAPITests::sendMonitoringFrames(const std::vector<data_conversion_layer::monitoring_frame::Message>& msgs)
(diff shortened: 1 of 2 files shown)


standalone/test/integration_tests/api/integrationtest_scanner_api.cpp
```diff
@@ -149,30 +155,94 @@ void ScannerAPITests::prepareScannerMockStartReply()
...
   strict_scanner_mock_->startListeningForControlMsg();
   const auto start_future = scanner_->start();
 
   ASSERT_TRUE(start_req_received_barrier->waitTillRelease(DEFAULT_TIMEOUT)) << "Start request not received";
-  ASSERT_EQ(start_future.wait_for(FUTURE_WAIT_TIMEOUT), std::future_status::timeout) << "Scanner::start() finished too "
-                                                                                        "early";
+  expectNotDoneYet("Scanner::start()", start_future);
   strict_scanner_mock_->sendStartReply();
-  ASSERT_EQ(start_future.wait_for(DEFAULT_TIMEOUT), std::future_status::ready) << "Scanner::start() not finished";
+  expectIsDone("Scanner::start()", start_future);
```
agutenkunst, line 245: Using a future allready enables concurrent usage. Why not just wait for `start_future` and remove the barrier?

standalone/test/integration_tests/api/integrationtest_scanner_api.cpp
```diff
@@ -181,14 +251,14 @@ TEST_F(ScannerAPITests, shouldReceiveStartRequestWithCorrectHostIpWhenUsingAutoI
   setUpScannerV2();
   setUpStrictScannerMock();
   const auto start_req_received_barrier =
-      prepareStrictMockStartRequestBarrier(generateScannerConfig(HOST_IP_ADDRESS, FRAGMENTED_SCAN));
+      prepareStartRequestBarrier(generateScannerConfig(HOST_IP_ADDRESS, FRAGMENTED_SCAN));
 
   strict_scanner_mock_->startListeningForControlMsg();
   const auto start_future = scanner_->start();
 
   ASSERT_TRUE(start_req_received_barrier->waitTillRelease(DEFAULT_TIMEOUT)) << "Start request not received";
```
agutenkunst, line 259: If you aim for readability I think lines like this are key. If I read this I have to think around a few corners. Actually I want to read that "Here the Mock is expected to receive a start request with some special config". Instead I read that some barrier is waiting to be release. Maybe this could be abstracted away.

standalone/test/integration_tests/api/integrationtest_scanner_api.cpp
```diff
@@ -149,30 +155,94 @@ void ScannerAPITests::prepareScannerMockStartReply()
...
+  const auto start_future = scanner_->start();
+  expectIsDone("Scanner::start()", start_future);
+}
+
+void ScannerAPITests::expectNotDoneYet(const std::string& method_name, const std::future<void>& future)
+{
+  EXPECT_EQ(future.wait_for(FUTURE_WAIT_TIMEOUT), std::future_status::timeout) << method_name << " finished to early";
+}
+
+void ScannerAPITests::expectIsDone(const std::string& method_name, const std::future<void>& future)
```
agutenkunst, line 210: No a fan. Returning a future is part of the API. Using exactly this future and its functions in the tests seems fine to me.

agutenkunst: Also passing the string just to be able to print it is not optimal. I immediatly assumed there is something more done with it.

rfeistenauer: You are totally right about the string. The goal was to make the intent of the expectation more clear, because expect future.wait_for() to be equal to something does not tell me what its about right away. I had to read the result string to get that. Introduced a define now that hopefully increases readability without passing a log string to a method

rfeistenauer: You are right about the string.
My intention was to ease the readability. I was always trying to understand the equation at specially on the = timeout part and then defaulted on reading the message on the line end.
Introduced a define now, which hopefully satisfies you complains and helps to read the line faster.

standalone/test/integration_tests/api/integrationtest_scanner_api.cpp
rfeistenauer: You are right, it would be nice if we could get that more clear. I would suggest to do that in a separate pr though

standalone/test/integration_tests/api/integrationtest_scanner_api.cpp
```diff
@@ -65,6 +65,10 @@ class UserCallbacks
   MOCK_METHOD1(LaserScanCallback, void(const LaserScan&));
 };
 
+#define EXPECT_FUTURE_IS_READY(val1) EXPECT_EQ(val1.wait_for(DEFAULT_TIMEOUT), std::future_status::ready)
+
+#define EXPECT_FUTURE_NOT_READY(val1) EXPECT_EQ(val1.wait_for(FUTURE_WAIT_TIMEOUT), std::future_status::timeout)
```
agutenkunst, line 70: Maybe put the Timeout as Parameter

agutenkunst:
```suggestion
#define EXPECT_FUTURE_TIMEOUT(val1) EXPECT_EQ(val1.wait_for(FUTURE_WAIT_TIMEOUT), std::future_status::timeout)
```

agutenkunst (approved):
Thank you!

PR #245, opened 2021-09-29
Minor improvement of TenthOfDegree

agutenkunst:
Removes the imho unnecessary member function `constexpr operator uint16_t() const` from `TenthOfDegree`.
Mathematical operations should talk place within the type as long as possible and convertion should be done explicitly at the last moment possible.

### DIFF
--- a/standalone/test/unit_tests/util/unittest_tenth_of_degree.cpp
+++ b/standalone/test/unit_tests/util/unittest_tenth_of_degree.cpp
@@ -15,6 +15,14 @@
 
 #include <gtest/gtest.h>
 
+#ifdef __linux__
+#define UNUSED __attribute__((unused))
+#endif
+
+#ifdef _WIN32
+#define UNUSED
+#endif
+
 #include "psen_scan_v2_standalone/data_conversion_layer/angle_conversions.h"
 #include "psen_scan_v2_standalone/util/tenth_of_degree.h"
 
@@ -143,9 +151,9 @@ TEST(TenthOfDegreeTest, LargerThanOrEqualComparison)
 
 TEST(TenthOfDegreeTest, shouldThrowUnderflowErrorOnCastToIntOfValuesUnderZero)
 {
-  EXPECT_THROW(std::cout << static_cast<uint16_t>(util::TenthOfDegree(-1))
-                         << " - printed to not be unused. Should throw before!\n",
-               std::underflow_error);
+  uint16_t result UNUSED;  // Macro needed to suppress clang-tidy warning
+  EXPECT_THROW(result = static_cast<uint16_t>(util::TenthOfDegree(-1)), std::underflow_error)
+      << "Should throw underflow error";
 }
 
 }  // namespace psen_scan_v2_standalone_test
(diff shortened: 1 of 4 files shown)


jschleicher (dismissed):

standalone/test/unit_tests/util/unittest_tenth_of_degree.cpp
```diff
@@ -135,9 +135,7 @@ TEST(TenthOfDegreeTest, LargerThanOrEqualComparison)
 
 TEST(TenthOfDegreeTest, shouldThrowUnderflowErrorOnCastToIntOfValuesUnderZero)
 {
-  EXPECT_THROW(std::cout << static_cast<uint16_t>(util::TenthOfDegree(-1))
-                         << " - printed to not be unused. Should throw before!\n",
-               std::underflow_error);
+  EXPECT_THROW(static_cast<uint16_t>(util::TenthOfDegree(-1)), std::underflow_error) << "Should throw underflow error";
```
jschleicher, line 138: Clang complains. Maybe the static_cast could be droped as well? 
```suggestion
  EXPECT_THROW(util::TenthOfDegree(-1), std::underflow_error) << "Should throw underflow error";
```
And in case the warning persists (see https://nelkinda.com/blog/suppress-warnings-in-gcc-and-clang/#w1223aab3b1c57):
```suggestion
  EXPECT_THROW(__attribute__((unused)) util::TenthOfDegree(-1), std::underflow_error) << "Should throw underflow error";
```

cdoehn, line 138: This should solve the issue
```suggestion
  uint16_t result __attribute__((unused));
  EXPECT_THROW(result = static_cast<uint16_t>(util::TenthOfDegree(-1)), std::underflow_error)
      << "Should throw underflow error";
```

agutenkunst: The solution needs to build with MSVC as well

agutenkunst: I think removing the cast removes a bit clarity about what is tested

standalone/test/unit_tests/util/unittest_tenth_of_degree.cpp
```diff
@@ -143,9 +151,9 @@ TEST(TenthOfDegreeTest, LargerThanOrEqualComparison)
 
 TEST(TenthOfDegreeTest, shouldThrowUnderflowErrorOnCastToIntOfValuesUnderZero)
 {
-  EXPECT_THROW(std::cout << static_cast<uint16_t>(util::TenthOfDegree(-1))
-                         << " - printed to not be unused. Should throw before!\n",
-               std::underflow_error);
+  uint16_t result UNUSED;  // Macro needed to suppress clang-tidy warning
+  EXPECT_THROW(result = static_cast<uint16_t>(util::TenthOfDegree(-1)), std::underflow_error)
```
jschleicher, line 155: is the cast necessary? Why not use the original type?
```suggestion
  EXPECT_THROW(result = util::TenthOfDegree(-1), std::underflow_error)
```

jschleicher (approved):
Maybe use the appropriate type without cast. Generally approving.

PR #253, opened 2021-10-08
Feature/active zone set

rfeistenauer:
## Description

These changes allow the user to read the active zoneset:
- The scanner is requested to always send the active zoneset now.
- The zoneset is added to the LaserScan msgs of the inner api.
- The active zoneset is published to ~/active_zoneset.

---

<details>

- [x] Public api function documentation
- [x] Package Readme ([example pilz_robots](https://github.com/PilzDE/pilz_robots/blob/melodic-devel/README.md))
- [x] CHANGELOG.rst updated
- [x] Copyright headers
- [x] ~Examples~

### Review Checklist

- [x] When is the new feature released? In the next weeks

### Hardware tests
- [x] Perform hardware tests

</details>

### DIFF
--- a/standalone/main.cpp
+++ b/standalone/main.cpp
@@ -35,7 +35,7 @@ const util::TenthOfDegree ANGLE_END{ data_conversion_layer::degreeToTenthDegree(
  */
 void laserScanCallback(const LaserScan& scan)
 {
-  // Available scan information is listed on
+  // Other data fields are listed on
   // https://docs.ros.org/en/melodic/api/psen_scan_v2/html/classpsen__scan__v2__standalone_1_1LaserScan.html
   PSENSCAN_INFO_THROTTLE(1 /* sec */, "laserScanCallback()", "Ranges {}", util::formatRange(scan.getMeasurements()));
 }


test/hw_tests/hwtest_zoneset_switching.cpp
```diff
@@ -0,0 +1,104 @@
...
+    }
+    else if (msg.data == 1 && received_active_zone_two.wait_for(NOT_SET_TIMEOUT) == std::future_status::timeout)
+    {
+      received_active_zone_two_promise.set_value();
+    }
+  };
+  auto zone_subscriber = nh.subscribe<std_msgs::UInt8>("/laser_1/active_zoneset", 10, callback);
+
+  sendRelayCmd(SCANNER_POWER_BIT);
+  EXPECT_FUTURE_IS_READY(received_active_zone_one, DEFAULT_TIMEOUT);
```
martiniil, line 87: I don't think it works like that. At least not deterministically. The relay cmd turns on the scanner (which it already is? Maybe we have to establish "rules" about which assumptions tests can make about the scanner status, if any.). Now you would have to wait until the scanner has booted and is started by the driver. Then you would have to wait for a first monitoring frame to be processed.

The `waitForMessage` call doesn't help you here. I am also curious about the callback implementation. I think you should stick to `async_barrier.h`.

test/include/psen_scan_v2/ros_integrationtest_helper.h
```diff
@@ -0,0 +1,72 @@
...
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU Lesser General Public License for more details.
+//
+// You should have received a copy of the GNU Lesser General Public License
+// along with this program.  If not, see <https://www.gnu.org/licenses/>.
+#ifndef PSEN_SCAN_V2_ROS_INTEGRATIONTEST_HELPER_H
+#define PSEN_SCAN_V2_ROS_INTEGRATIONTEST_HELPER_H
+
+#include <chrono>
+#include <iostream>  // TODO remove
```
martiniil, line 19:
```suggestion
#include <string>
#include <sstream>
#include <thread>

#include <gtest/gtest.h>
```

test/include/psen_scan_v2/ros_integrationtest_helper.h
```diff
@@ -0,0 +1,72 @@
...
+
+#include <ros/master.h>
+
+using namespace std::chrono_literals;
+
+namespace psen_scan_v2_test
+{
+::testing::AssertionResult TopicExists(const std::string& topic,
+                                       const std::chrono::milliseconds& timeout = 500ms,
+                                       size_t retries = 10)
```
martiniil, line 29: Maybe rename `timeout` to `sleep_time` or s.th. similar.

test/hw_tests/hwtest_zoneset_switching.cpp
rfeistenauer: I only want to make sure, that the scanner switches to zoneset 0. I expect the scanner to be powered and don't turn him off in this test. I introduced a change, that hopefully makes this more clear to the reader. The barrier to my understanding would not block the scanner any more than the wait for future does in this case. If the scanner is disabled and just starts with this command due to an unexpected test precondition, then the test will fail but that is true for every hw test we have so far isn't it?

PR #272, opened 2021-11-03
Avoid throwing exception that cannot be catched by user

### LINKED ISSUE #271: Exception from the node not caught
The exception  “Unknown operation result code” raised by the driver are not catched by the ROS wrapper and causing the node to terminate immediately. 

https://github.com/PilzDE/psen_scan_v2/blob/d3b2405e3edf134e76123471a84b4ede8e3050e6/src/psen_scan_driver.cpp#L118 `std::future.wait()` doesn't retrieve the result and thus lets the exception unhandled. To properly make use of the try...catch contruction, there should be a call that handles the result of the thread.

While the return code should still be negative in case of failures, it is desirable, to get the ROS_ERROR from https://github.com/PilzDE/psen_scan_v2/blob/d3b2405e3edf134e76123471a84b4ede8e3050e6/src/psen_scan_driver.cpp#L120-L122 instead of just terminating.

psen-scan-throw

martiniil:
## Description

These changes Fix #271

---

<details>

- [ ] Public api function documentation
- [ ] Architecture documentation reflects actual code structure
- [ ] [Tutorials](https://wiki.ros.org/pilz_robots/Tutorials/)
- [ ] Overview on [ROS wiki](https://wiki.ros.org/pilz_robots)
- [ ] Package Readme ([example pilz_robots](https://github.com/PilzDE/pilz_robots/blob/melodic-devel/README.md))
- [ ] Good commit messages ([some tips](https://dev.to/jacobherrington/how-to-write-useful-commit-messages-my-commit-message-template-20n9))
- [ ] CHANGELOG.rst updated
- [ ] Copyright headers
- [ ] Examples

### Review Checklist
- [ ] Soft- and hardware architecture (diagrams and description)
- [ ] Test review (test plan and individual test cases)
- [ ] Documentation describes purpose of file(s) and responsibilities
- [ ] Code (coding rules, style guide)

- [ ] When is the new feature released?
- [ ] Which dependent packages do have to be released simultaneously?

### Hardware tests
- [ ] ~Perform hardware tests~

</details>

### DIFF
--- a/standalone/test/include/psen_scan_v2_standalone/communication_layer/scanner_mock.h
+++ b/standalone/test/include/psen_scan_v2_standalone/communication_layer/scanner_mock.h
@@ -76,7 +76,8 @@ public:
                void(const udp::endpoint&, const psen_scan_v2_standalone::data_conversion_layer::RawData&));
 
 public:
-  void sendStartReply();
+  void sendStartReply(const data_conversion_layer::scanner_reply::Message::OperationResult& result =
+                          data_conversion_layer::scanner_reply::Message::OperationResult::accepted);
   void sendStopReply();
   void sendMonitoringFrame(const data_conversion_layer::monitoring_frame::Message& msg);
   void sendEmptyMonitoringFrame();
@@ -84,7 +85,9 @@ public:
 
 private:
   void startContinuousListeningForControlMsg();
-  void sendReply(const data_conversion_layer::scanner_reply::Message::Type& reply_type);
+  void sendReply(const data_conversion_layer::scanner_reply::Message::Type& reply_type,
+                 const data_conversion_layer::scanner_reply::Message::OperationResult& result =
+                     data_conversion_layer::scanner_reply::Message::OperationResult::accepted);
 
 private:
   const udp::endpoint control_msg_receiver_;
--- a/standalone/test/integration_tests/api/integrationtest_scanner_api.cpp
+++ b/standalone/test/integration_tests/api/integrationtest_scanner_api.cpp
@@ -75,6 +75,12 @@ public:
       receiveControlMsg(                                                                                               \
           _, data_conversion_layer::start_request::serialize(data_conversion_layer::start_request::Message(config))))
 
+#define ON_START_REQUEST_CALL(hw_mock, config)                                                                         \
+  ON_CALL(                                                                                                             \
+      hw_mock,                                                                                                         \
+      receiveControlMsg(                                                                                               \
+          _, data_conversion_layer::start_request::serialize(data_conversion_layer::start_request::Message(config))))
+
 #define EXPECT_CALLBACK_WILL_OPEN_BARRIER(cb, msgs, barrier)                                                           \
   do                                                                                                                   \
   {                                                                                                                    \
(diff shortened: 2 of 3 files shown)


standalone/test/include/psen_scan_v2_standalone/communication_layer/scanner_mock.h
```diff
@@ -76,15 +76,18 @@ class ScannerMock
...
   void sendStopReply();
   void sendMonitoringFrame(const data_conversion_layer::monitoring_frame::Message& msg);
   void sendEmptyMonitoringFrame();
   void sendMonitoringFrames(const std::vector<data_conversion_layer::monitoring_frame::Message>& msgs);
 
 private:
   void startContinuousListeningForControlMsg();
-  void sendReply(const data_conversion_layer::scanner_reply::Message::Type& reply_type);
+  void sendReply(const data_conversion_layer::scanner_reply::Message::Type& reply_type,
+                 const data_conversion_layer::scanner_reply::Message::OperationResult& result =
```
martiniil, line 89: using ReplyMessage = ...?

martiniil: @martiniil todo: extend integrationtest_ros_scanner_node

test/integration_tests/integrationtest_ros_scanner_node.cpp
```diff
@@ -173,6 +173,40 @@ TEST_F(RosScannerNodeTests, shouldStartAndStopSuccessfullyIfScannerRespondsToReq
...
+      nh_priv_, "scan", "scanner", configuration::DEFAULT_X_AXIS_ROTATION, scanner_config_);
+
+  util::Barrier start_barrier;
+  std::promise<void> hw_finished_request;
+  ON_CALL(ros_scanner_node.scanner_, start())
+      .WillByDefault(DoAll(OpenBarrier(&start_barrier), Return_Future(hw_finished_request)));
+
+  std::future<void> loop = std::async(std::launch::async, [&ros_scanner_node]() { ros_scanner_node.run(); });
+  start_barrier.waitTillRelease(DEFAULT_TIMEOUT);
+  std::this_thread::sleep_for(1s);
```
rfeistenauer, line 201: 1. this is a ros test, you could use ros::Duration().sleep() 
2. if you dont try to do a loop.get() before the sleep, how do you know it is not set with an exception already? What does the sleep do here?

martiniil: 1. thanks
2. I know what is set and when because I (the test thread) own the promise. What I do not know is how `scanner.run()` reacts if I delay setting the exception. It might not wait. So I delay setting the exception and expect that `scanner.run()` still throws.

martiniil: btw This is exactly what happens in "real life". After `scanner.start()` is called some time passes until the scanner reply arrives which leads to an exception.

rfeistenauer (approved):

standalone/include/psen_scan_v2_standalone/protocol_layer/scanner_state_machine.h
```diff
@@ -223,10 +234,15 @@ class ScannerProtocolDef : public msm::front::state_machine_def<ScannerProtocolD
     InternalScannerReplyError(const std::string& error_msg);
   };
   // LCOV_EXCL_STOP
-  void checkForInternalErrors(const data_conversion_layer::scanner_reply::Message& msg);
+  bool isStartReply(data_conversion_layer::scanner_reply::Message const& msg);
+  bool isAcceptedReply(data_conversion_layer::scanner_reply::Message const& msg);
+  bool isUnknownReply(data_conversion_layer::scanner_reply::Message const& msg);
+  bool isRefusedReply(data_conversion_layer::scanner_reply::Message const& msg);
 
+  void checkForInternalErrors(const data_conversion_layer::scanner_reply::Message& msg);
```
jschleicher, line 242: Is this function still used?

jschleicher (commented):
See nitpick comments inline.

standalone/include/psen_scan_v2_standalone/protocol_layer/scanner_state_machine.h
martiniil: Yes.

Issue #262, opened 2021-10-25
Flaky integrationtest_scanner_api


martiniil:
Keep track of all failures occuring recently.

<details>
<summary>LaserScanShouldContainAllInfosTransferedByMonitoringFrameMsg</summary>

https://github.com/PilzDE/psen_scan_v2/runs/3994840895?check_suite_focus=true
```
2021-10-25T09:32:41.3655140Z     <testcase name="LaserScanShouldContainAllInfosTransferedByMonitoringFrameMsg" status="run" result="completed" time="1.003" timestamp="2021-10-25T09:32:38" classname="ScannerAPITests">
2021-10-25T09:32:43.7392993Z       <failure message="/root/target_ws/src/psen_scan_v2/standalone/test/integration_tests/api/integrationtest_scanner_api.cpp:359&#x0A;Expected equality of these values:&#x0A;  future.wait_for(std::chrono::seconds{ 1 })&#x0A;    Which is: 4-byte object &lt;01-00 00-00&gt;&#x0A;  std::future_status::ready&#x0A;    Which is: 4-byte object &lt;00-00 00-00&gt;&#x0A;stop_future = driver_-&gt;stop(); does not return." type=""><![CDATA[/root/target_ws/src/psen_scan_v2/standalone/test/integration_tests/api/integrationtest_scanner_api.cpp:359
2021-10-25T09:32:43.7399968Z Expected equality of these values:
2021-10-25T09:32:43.7416691Z   future.wait_for(std::chrono::seconds{ 1 })
2021-10-25T09:32:43.7418028Z     Which is: 4-byte object <01-00 00-00>
2021-10-25T09:32:43.7418530Z   std::future_status::ready
2021-10-25T09:32:43.7419202Z     Which is: 4-byte object <00-00 00-00>
2021-10-25T09:32:43.7419964Z stop_future = driver_->stop(); does not return.]]></failure>
2021-10-25T09:32:43.7420522Z     </testcase>
```
</details>

<details>
<summary>shouldNotCallLaserscanCallbackInCaseOfEmptyMonitoringFrame</summary>

https://github.com/PilzDE/psen_scan_v2/runs/3996672390?check_suite_focus=true
```
2021-10-25T12:25:51.1925992Z     <testcase name="shouldNotCallLaserscanCallbackInCaseOfEmptyMonitoringFrame" status="run" time="1.003" classname="ScannerAPITests">
[...]

PR #279, opened 2021-11-17
Refactor/integrationtest scanner api

rfeistenauer:
## Description

This test aims to improve the integration test readability and reduce its complexity by:
* Using Fixtures for common test setup
* Removing barrier opens expectations (they get opened by expect calls, so there is an expectation already)
* Use chrono literals instead of DEFAULT_TIMEOUT since they are more descriptive
* some smaller cleanups

FYI: I also tried to get rid of the EXPECT_SCANNER_TO_STOP_SUCCESSFULLY and the respective start expectation but:
* We are using a strict mock for the scanner so any call has to be EXPECTed => ON_CALL is no option and there is no ASSERT_CALL 
* we don't want to hide them away so we need to keep them in the tests itself.
* Removing the stop or start completely and expecting everything to still work is unreasonable
* Changing the ScanerMock to just do some actions non asynchrounouse to make the barriers unneccessary and therefore the expectations is at least not easy and maybe not a good idea at all.

---

<details>

- [x] ~CHANGELOG.rst updated~
- [x] ~Copyright headers~
- [x] ~Examples~

### Review Checklist
- [x] ~Documentation describes purpose of file(s) and responsibilities~

### Hardware tests
- [x] ~Perform hardware tests~

</details>

### DIFF
--- a/standalone/test/integration_tests/api/integrationtest_scanner_api.cpp
+++ b/standalone/test/integration_tests/api/integrationtest_scanner_api.cpp
@@ -163,11 +163,20 @@ void ScannerAPITests::setUpScannerHwMock()
   hw_mock_.reset(new StrictMock<ScannerMock>{ HOST_IP_ADDRESS, port_holder_ });
 }
 
-TEST_F(ScannerAPITests, shouldSendStartRequestAndReturnValidFutureWhenLaunchingWithValidConfig)
+class ScannerAPITestsFragmented : public ScannerAPITests
 {
-  setUpScannerConfig();
-  setUpScannerV2Driver();
-  setUpScannerHwMock();
+  void SetUp() override
+  {
+    port_holder_.printPorts();
+    setLogLevel(CONSOLE_BRIDGE_LOG_DEBUG);
+    setUpScannerConfig();
+    setUpScannerV2Driver();
+    setUpScannerHwMock();
+  }
+};
+
+TEST_F(ScannerAPITestsFragmented, shouldSendStartRequestAndReturnValidFutureWhenLaunchingWithValidConfig)
+{
   util::Barrier start_req_received_barrier;
   EXPECT_START_REQUEST_CALL(*hw_mock_, *config_).WillOnce(OpenBarrier(&start_req_received_barrier));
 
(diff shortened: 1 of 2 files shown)


standalone/test/integration_tests/api/integrationtest_scanner_api.cpp
```diff
@@ -92,9 +91,9 @@ class UserCallbacks
     EXPECT_START_REQUEST_CALL(*hw_mock, *config).WillOnce(OpenBarrier(&start_req_barrier));                            \
     EXPECT_NO_BLOCK_NO_THROW(start_future = driver->start(););                                                         \
     EXPECT_TRUE(start_future.valid());                                                                                 \
-    EXPECT_BARRIER_OPENS(start_req_barrier, DEFAULT_TIMEOUT) << "Start request not received";                          \
+    start_req_barrier.waitTillRelease(2s);                                                                             \
```
martiniil, line 94: Is there a specific reason for decreasing the timeout (was 3s before)? Keep in mind we still have #262 open. These failures might not have occured for a bit looser timeouts.

standalone/test/integration_tests/api/integrationtest_scanner_api.cpp
```diff
@@ -163,21 +162,40 @@ void ScannerAPITests::setUpScannerHwMock()
...
+  void SetUp() override
+  {
+    ScannerAPITests::SetUp();
+    setUpScannerConfig(HOST_IP_ADDRESS, UNFRAGMENTED_SCAN);
+    setUpScannerV2Driver();
+    setUpScannerHwMock();
+  }
+};
+
+TEST_F(ScannerAPITestsFragmented, shouldSendStartRequestAndReturnValidFutureWhenLaunchingWithValidConfig)
```
martiniil, line 187: For this test (and probably a few others) the fragmented property is not relevant.

I suggest to use the `ScannerAPITests` fixture for these test cases and also setup a default config in `ScannerAPITests.SetUp()`. It will get overwritten in the child classes anyways.

martiniil: Eventually, we can decide if it makes sense to sort the tests according to their fixture. Or even separate the integrationtest into multiple files with an own header for the common fixture.

martiniil (changes requested):
See inline comments. It is nice that we gradually improve the readability. :+1:

standalone/test/integration_tests/api/integrationtest_scanner_api.cpp
rfeistenauer: the #262 runs against a different 1s timeout in a method that should not even take .1 seconds.
The whole test cases are usually done in far less that one second. I think high timeouts only resolve in costing us more time during development with the risk of covering other problems. If it wasn't for the UDP in those tests i would even pledge for timeouts of less than a second. I thought going down to 2s will reduce the time we wait for the CI in fail cases by 30% without really risking to cause problems due to to small timeouts.

standalone/test/integration_tests/api/integrationtest_scanner_api.cpp
rfeistenauer: either the fragmented or the unfragmented property has to be the defautl case. up to now it was the fragmented one. switching to ScannerApiTests even with adding the default config there, would resault in having to create driver and mock objects again. Or if you add those creations to the ScannerAPITests.SetUp, you will have problems with "shouldThrowWhenConstructedWithInvalidLaserScanCallback" which does set up the scanner itself and uses no hw_mock.

martiniil (approved):

PR #317, opened 2022-01-28
Add author and urls to package.xml

agutenkunst:
## Description
* Adds urls
* Fixes missing field in http://wiki.ros.org/psen_scan_v2

missing_author

### DIFF
--- a/package.xml
+++ b/package.xml
@@ -3,7 +3,7 @@
   <name>psen_scan_v2</name>
   <version>0.10.1</version>
   <description>ROS support for the Pilz laser scanner</description>
-  
+
   <maintainer email="a.gutenkunst@pilz.de">Alexander Gutenkunst</maintainer>
   <maintainer email="c.doehn@pilz.de">Christian Döhn</maintainer>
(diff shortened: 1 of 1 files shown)


package.xml
```diff
@@ -3,14 +3,22 @@
...
+  
   <maintainer email="a.gutenkunst@pilz.de">Alexander Gutenkunst</maintainer>
   <maintainer email="c.doehn@pilz.de">Christian Döhn</maintainer>
   <maintainer email="c.henkel@pilz.de">Christian Henkel</maintainer>
   <maintainer email="i.martini@pilz.de">Immanuel Martini</maintainer>
   <maintainer email="r.feistenauer@pilz.de">Richard Feistenauer</maintainer>
 
   <license>LGPLv3</license>
+  
+  <url type="website">https://wiki.ros.org/psen_scan_v2</url>
```
martiniil, line 15: As far as I see this will display as "External Website", but will lead to the site you are already on. I would skip it.

martiniil: Or link to https://www.pilz.com/en-INT/products/robotics/ros-modules ?

martiniil (approved):

